PR #6, opened 2025-01-30
`CreateMint` instruction

grod220:
This PR builds out the `CreateMint` instruction enabling the creation of "wrapped" token-2022 tokens from existing spl-token tokens (and vice versa). The intention is to facilitate interoperability between different token standards.

Mainly creates two accounts:

*   **Wrapped Mint**  Provisions a new mint account for the wrapped token. The address of this wrapped mint is deterministically derived as a PDA from the unwrapped mint address.
*   **Backpointer**  Provisions a backpointer account (PDA), derived from the wrapped mint address. This account stores a reference to the original unwrapped mint, facilitating reverse lookups.

Future work to build out remaining wrap & unwrap instructions.

### DIFF
--- a/program/src/instruction.rs
+++ b/program/src/instruction.rs
@@ -48,12 +48,10 @@ pub enum TokenWrapInstruction {
     ///    `get_wrapped_mint_address(unwrapped_mint_address,
     ///    wrapped_token_program_id)`
     /// 4. `[writeable]` Recipient wrapped token account
-    /// 5. `[]` Escrow mint authority, address must be:
-    ///    `get_wrapped_mint_authority(wrapped_mint)`
-    /// 6. `[]` SPL Token program for unwrapped mint
-    /// 7. `[]` SPL Token program for wrapped mint
-    /// 8. `[signer]` Transfer authority on unwrapped token account
-    /// 9. `..8+M` `[signer]` (Optional) M multisig signers on unwrapped token
+    /// 5. `[]` SPL Token program for unwrapped mint
+    /// 6. `[]` SPL Token program for wrapped mint
+    /// 7. `[signer]` Transfer authority on unwrapped token account
+    /// 8. `..8+M` `[signer]` (Optional) M multisig signers on unwrapped token
     ///    account
     Wrap {
         /// little-endian `u64` representing the amount to wrap
@@ -75,12 +73,10 @@ pub enum TokenWrapInstruction {
     ///    `get_wrapped_mint_authority(wrapped_mint_address)`
     /// 3. `[writeable]` Recipient unwrapped tokens
     /// 4. `[]` Unwrapped token mint
-    /// 5. `[]` Escrow unwrapped token authority
-    ///    `get_wrapped_mint_authority(wrapped_mint)`
-    /// 6. `[]` SPL Token program for wrapped mint
-    /// 7. `[]` SPL Token program for unwrapped mint
-    /// 8. `[signer]` Transfer authority on wrapped token account
-    /// 9. `..8+M` `[signer]` (Optional) M multisig signers on wrapped token
+    /// 5. `[]` SPL Token program for wrapped mint
+    /// 6. `[]` SPL Token program for unwrapped mint
+    /// 7. `[signer]` Transfer authority on wrapped token account
+    /// 8. `..8+M` `[signer]` (Optional) M multisig signers on wrapped token
     ///    account
     Unwrap {
         /// little-endian `u64` representing the amount to unwrap
@@ -156,44 +152,6 @@ pub fn create_mint(
     Instruction::new_with_bytes(*program_id, &data, accounts)
 }
 
-/// Creates `UnWrap` instruction.
-#[allow(clippy::too_many_arguments)]
-pub fn unwrap(
-    program_id: &Pubkey,
-    wrapped_token_account_address: &Pubkey,
-    wrapped_mint_address: &Pubkey,
-    unwrapped_escrow_address: &Pubkey,
-    recipient_unwrapped_token_account_address: &Pubkey,
-    unwrapped_mint_address: &Pubkey,
-    wrapped_mint_authority_address: &Pubkey,
-    wrapped_token_program_id: &Pubkey,
-    unwrapped_token_program_id: &Pubkey,
-    transfer_authority_address: &Pubkey,
-    multisig_signer_pubkeys: &[&Pubkey],
-    amount: u64,
-) -> Instruction {
-    let mut accounts = vec![
-        AccountMeta::new(*wrapped_token_account_address, false),
-        AccountMeta::new(*wrapped_mint_address, false),
-        AccountMeta::new(*unwrapped_escrow_address, false),
-        AccountMeta::new(*recipient_unwrapped_token_account_address, false),
-        AccountMeta::new_readonly(*unwrapped_mint_address, false),
-        AccountMeta::new_readonly(*wrapped_mint_authority_address, false),
-        AccountMeta::new_readonly(*wrapped_token_program_id, false),
-        AccountMeta::new_readonly(*unwrapped_token_program_id, false),
-        AccountMeta::new_readonly(
-            *transfer_authority_address,
-            multisig_signer_pubkeys.is_empty(),
-        ),
-    ];
-    for signer_pubkey in multisig_signer_pubkeys.iter() {
-        accounts.push(AccountMeta::new_readonly(**signer_pubkey, true));
-    }
-
-    let data = TokenWrapInstruction::Unwrap { amount }.pack();
-    Instruction::new_with_bytes(*program_id, &data, accounts)
-}
-
 /// Creates `Wrap` instruction.
 #[allow(clippy::too_many_arguments)]
 pub fn wrap(
@@ -203,7 +161,6 @@ pub fn wrap(
     unwrapped_mint_address: &Pubkey,
     wrapped_mint_address: &Pubkey,
     recipient_wrapped_token_account_address: &Pubkey,
-    wrapped_mint_authority_address: &Pubkey,
     unwrapped_token_program_id: &Pubkey,
     wrapped_token_program_id: &Pubkey,
     transfer_authority_address: &Pubkey,
@@ -216,7 +173,6 @@ pub fn wrap(
         AccountMeta::new_readonly(*unwrapped_mint_address, false),
         AccountMeta::new(*wrapped_mint_address, false),
         AccountMeta::new(*recipient_wrapped_token_account_address, false),
-        AccountMeta::new_readonly(*wrapped_mint_authority_address, false),
         AccountMeta::new_readonly(*unwrapped_token_program_id, false),
         AccountMeta::new_readonly(*wrapped_token_program_id, false),
         AccountMeta::new_readonly(
@@ -229,5 +185,41 @@ pub fn wrap(
     }
 
     let data = TokenWrapInstruction::Wrap { amount }.pack();
+    Instruction::new_with_bytes(*program_id, &data, accounts)
+}
+
+/// Creates `Unwrap` instruction.
+#[allow(clippy::too_many_arguments)]
+pub fn unwrap(
+    program_id: &Pubkey,
+    wrapped_token_account_address: &Pubkey,
+    wrapped_mint_address: &Pubkey,
+    unwrapped_escrow_address: &Pubkey,
+    recipient_unwrapped_token_account_address: &Pubkey,
+    unwrapped_mint_address: &Pubkey,
+    wrapped_token_program_id: &Pubkey,
+    unwrapped_token_program_id: &Pubkey,
+    transfer_authority_address: &Pubkey,
+    multisig_signer_pubkeys: &[&Pubkey],
+    amount: u64,
+) -> Instruction {
+    let mut accounts = vec![
+        AccountMeta::new(*wrapped_token_account_address, false),
+        AccountMeta::new(*wrapped_mint_address, false),
+        AccountMeta::new(*unwrapped_escrow_address, false),
+        AccountMeta::new(*recipient_unwrapped_token_account_address, false),
+        AccountMeta::new_readonly(*unwrapped_mint_address, false),
+        AccountMeta::new_readonly(*wrapped_token_program_id, false),
+        AccountMeta::new_readonly(*unwrapped_token_program_id, false),
+        AccountMeta::new_readonly(
+            *transfer_authority_address,
+            multisig_signer_pubkeys.is_empty(),
+        ),
+    ];
+    for signer_pubkey in multisig_signer_pubkeys.iter() {
+        accounts.push(AccountMeta::new_readonly(**signer_pubkey, true));
+    }
+
+    let data = TokenWrapInstruction::Unwrap { amount }.pack();
     Instruction::new_with_bytes(*program_id, &data, accounts)
 }
--- a/program/src/processor.rs
+++ b/program/src/processor.rs
@@ -64,11 +64,11 @@ pub fn process_create_mint(
         }
         if wrapped_mint_account.owner != wrapped_token_program_account.key {
             msg!("Wrapped mint account owner is not the expected token program");
-            return Err(ProgramError::InvalidAccountData);
+            return Err(ProgramError::InvalidAccountOwner);
         }
         if wrapped_backpointer_account.owner != program_id {
             msg!("Wrapped backpointer account owner is not the expected token wrap program");
-            return Err(ProgramError::InvalidAccountData);
+            return Err(ProgramError::InvalidAccountOwner);
         }
         return Ok(());
     }
@@ -90,7 +90,7 @@ pub fn process_create_mint(
             "Error: wrapped_mint_account requires pre-funding of {} lamports",
             mint_rent_required
         );
-        return Err(ProgramError::InsufficientFunds);
+        return Err(ProgramError::AccountNotRentExempt);
     }
 
     // Initialize the wrapped mint
@@ -180,7 +180,7 @@ pub fn process_instruction(
             unimplemented!();
         }
         TokenWrapInstruction::Unwrap { .. } => {
-            msg!("Instruction: UnWrap");
+            msg!("Instruction: Unwrap");
             unimplemented!();
         }
     }
--- a/program/tests/test_create_mint.rs
+++ b/program/tests/test_create_mint.rs
@@ -192,7 +192,7 @@ fn test_idempotency_true_with_existing_invalid_accounts() {
     mollusk.process_and_validate_instruction(
         &instruction,
         accounts,
-        &[Check::err(ProgramError::InvalidAccountData)],
+        &[Check::err(ProgramError::InvalidAccountOwner)],
     );
 
     // Incorrect owner on wrapped backpointer account
@@ -226,7 +226,7 @@ fn test_idempotency_true_with_existing_invalid_accounts() {
     mollusk.process_and_validate_instruction(
         &instruction,
         accounts,
-        &[Check::err(ProgramError::InvalidAccountData)],
+        &[Check::err(ProgramError::InvalidAccountOwner)],
     );
 }
 
@@ -273,7 +273,7 @@ fn test_create_mint_insufficient_funds() {
     mollusk.process_and_validate_instruction(
         &instruction,
         accounts,
-        &[Check::err(ProgramError::InsufficientFunds)],
+        &[Check::err(ProgramError::AccountNotRentExempt)],
     );
 }
 


program/src/processor.rs
```diff
@@ -1,25 +1,168 @@
...
+    program_id: &Pubkey,
+    accounts: &[AccountInfo],
+    idempotent: bool,
+) -> ProgramResult {
+    let account_info_iter = &mut accounts.iter();
+
+    let wrapped_mint_account = next_account_info(account_info_iter)?;
+    let wrapped_backpointer_account = next_account_info(account_info_iter)?;
+    let unwrapped_mint_account = next_account_info(account_info_iter)?;
+    let _system_program_account = next_account_info(account_info_iter)?;
```
grod220, line 33: Conceptually not sure why a non-used account needs to be passed when it's not used in the code. However, tests indeed throw without it.

program/src/processor.rs
```diff
@@ -1,25 +1,168 @@
...
+    let decimals = unpacked_unwrapped_mint.decimals;
+
+    let (wrapped_mint_authority, authority_bump_seed) =
+        get_wrapped_mint_authority_with_seed(wrapped_mint_account.key);
+    let authority_bump_seeds = [authority_bump_seed];
+    let authority_signer_seeds =
+        _get_wrapped_mint_authority_signer_seeds(wrapped_mint_account.key, &authority_bump_seeds);
+
+    invoke_signed(
+        &initialize_mint2(
```
grod220, line 96: Not sure why there is mint & mint2 for this. Went with the later version though.

program/tests/test_create_mint.rs
```diff
@@ -0,0 +1,440 @@
...
+    assert!(wrapped_mint_data.is_initialized);
+    assert!(wrapped_mint_data.freeze_authority.is_none());
+
+    // Assert state of resulting backpointer account
+
+    let resulting_backpointer_account = &result.resulting_accounts[1].1;
+    assert_eq!(resulting_backpointer_account.owner, program_id);
+
+    let backpointer = bytemuck::from_bytes::<Backpointer>(&resulting_backpointer_account.data[..]);
+    assert_eq!(backpointer.unwrapped_mint, unwrapped_mint_address);
```
grod220, line 353: Curious to hear reviewers feedback on any other conditions to assert after the transaction.

buffalojoec:
@grod220 as per discussion on Zoom, there were a few small changes you wanted to make, then I can review. Just let me know when you're ready with a comment + tag!

grod220:
@buffalojoec Ready for your review 🙏

program/src/processor.rs
buffalojoec: Also another thing we chatted about offline, but the original `initialize_mint` required the account info for the `Rent` sysvar. After some time, the `Rent` sysvar became available via syscall, no longer requiring the account to be passed in. This instruction allows clients to invoke `InitializeMint` without providing the rent sysvar, while maintaining maximum backwards compatibility.
https://github.com/anza-xyz/agave/blob/ec86c958acd9e2983e8f3197ec2cb3a18247d590/programs/bpf_loader/src/syscalls/sysvar.rs#L122

program/src/processor.rs
```diff
@@ -1,25 +1,169 @@
...
+        unwrapped_mint_account.key,
+        wrapped_token_program_account.key,
+    );
+    let bump_seed = [bump];
+    let signer_seeds = _get_wrapped_mint_signer_seeds(
+        unwrapped_mint_account.key,
+        wrapped_token_program_account.key,
+        &bump_seed,
+    );
+    let space = spl_token_2022::state::Mint::get_packed_len();
```
buffalojoec, line 59: This is okay if we are assuming this program will not allow any Token-2022 extensions. If it does, we'll need to adjust this flow here. However, I think for simplicity's sake we shouldn't add support for extensions until absolutely necessary.

program/tests/test_create_mint.rs
```diff
@@ -0,0 +1,450 @@
...
+        false,
+    );
+
+    // Simulating existing data on mint or backpointer
+    let account_with_data = Account {
+        data: vec![1; 10],
+        ..Account::default()
+    };
+
+    // idempotent: true causes these to fail
```
buffalojoec, line 90:
```suggestion
    // idempotent: false causes these to fail
```

program/src/processor.rs
```diff
@@ -1,25 +1,169 @@
...
+    let wrapped_token_program_account = next_account_info(account_info_iter)?;
+
+    // Idempotency checks
+
+    if wrapped_mint_account.data_len() > 0 || wrapped_backpointer_account.data_len() > 0 {
+        msg!("Wrapped mint or backpointer account already initialized");
+        return if !idempotent {
+            Err(ProgramError::AccountAlreadyInitialized)
+        } else {
+            Ok(())
```
buffalojoec, line 43: This returns `Ok(())` if the account data is already initialized, no matter who the account's owner is, etc.

https://github.com/solana-program/associated-token-account/blob/1f3467ed3305e7b7dd0b7a34db0daf297366914d/program/src/instruction.rs#L20-L21

buffalojoec (commented):
Looking great so far! Nice work.

I left a bunch of comments, a lot of them small/nits, and I only skimmed the tests for now. I'll look at those closer once we've chatted about the other items!

program/src/processor.rs
```diff
@@ -1,25 +1,169 @@
...
+        &system_instruction::assign(&wrapped_mint_address, wrapped_token_program_account.key),
+        &[wrapped_mint_account.clone()],
+        &[&signer_seeds],
+    )?;
+
+    // New wrapped mint matches decimals & freeze authority of unwrapped mint
+    let unwrapped_mint_data = unwrapped_mint_account.try_borrow_data()?;
+    let unpacked_unwrapped_mint = spl_token_2022::state::Mint::unpack(&unwrapped_mint_data)?;
+    let decimals = unpacked_unwrapped_mint.decimals;
+    let freeze_authority = unpacked_unwrapped_mint.freeze_authority.as_ref().into();
```
joncinque, line 88: This actually raises a very good question... should the freeze authority for the unwrapped mint be able to freeze wrapped tokens? My thinking is to not have any freeze authority

program/src/processor.rs
```diff
@@ -1,25 +1,169 @@
...
+    )?;
+    invoke_signed(
+        &system_instruction::assign(&wrapped_mint_address, wrapped_token_program_account.key),
+        &[wrapped_mint_account.clone()],
+        &[&signer_seeds],
+    )?;
+
+    // New wrapped mint matches decimals & freeze authority of unwrapped mint
+    let unwrapped_mint_data = unwrapped_mint_account.try_borrow_data()?;
+    let unpacked_unwrapped_mint = spl_token_2022::state::Mint::unpack(&unwrapped_mint_data)?;
```
joncinque, line 86: Since the unwrapped mint might be a token-2022 mint with extensions, `unpack` will fail -- you'll need to use `StateWithExtensions::unpack`, ie https://github.com/solana-program/token-2022/blob/f69cc989c2a3f2657a6ad6041018468807223f08/program/src/extension/mod.rs#L1655

Bonus if you use the Pod variants, since they're zero-copy :grin:

program/tests/test_create_mint.rs
```diff
@@ -0,0 +1,450 @@
...
+};
+use {
+    mollusk_svm::{result::Check, Mollusk},
+    solana_program::pubkey::Pubkey,
+    spl_token_wrap::instruction::create_mint,
+};
+
+const MINT_DECIMALS: u8 = 12;
+const MINT_SUPPLY: u64 = 500_000_000;
+const FREEZE_AUTHORITY: &str = "11111115q4EpJaTXAZWpCg3J2zppWGSZ46KXozzo9";
```
joncinque, line 20: nit: may as well initialize this as a const Pubkey rather than using `from_str_const` for every test

program/tests/test_create_mint.rs
```diff
@@ -0,0 +1,450 @@
...
+
+    let lamports = rent.minimum_balance(data.len());
+
+    Account {
+        lamports,
+        data,
+        owner: spl_token_2022::id(),
+        ..Default::default()
+    }
+}
```
joncinque, line 62: nit: spl_token and spl_token_2022's mints are exactly the same, so you can combine both of these functions by also taking the `owner` program id

program/tests/test_create_mint.rs
```diff
@@ -0,0 +1,450 @@
...
+
+#[test]
+fn test_idempotency_false_with_existing_account() {
+    let program_id = spl_token_wrap::id();
+    let mut mollusk = Mollusk::new(&program_id, "spl_token_wrap");
+    mollusk_svm_programs_token::token2022::add_program(&mut mollusk);
+
+    let wrapped_mint_account = Pubkey::new_unique();
+    let wrapped_backpointer_account = Pubkey::new_unique();
```
joncinque, line 71: nit: this isn't too important because the test fails earlier, but if the logic ever gets moved, these will fail because of improper derivation, and it'll be annoying to debug. Can these be proper PDAs?

program/tests/test_create_mint.rs
```diff
@@ -0,0 +1,450 @@
...
+}
+
+#[test]
+fn test_idempotency_true_with_existing_account() {
+    let program_id = spl_token_wrap::id();
+    let mut mollusk = Mollusk::new(&program_id, "spl_token_wrap");
+    mollusk_svm_programs_token::token2022::add_program(&mut mollusk);
+
+    let wrapped_mint_account = Pubkey::new_unique();
+    let wrapped_backpointer_account = Pubkey::new_unique();
```
joncinque, line 125: nit: same with these as above, let's properly derive these

program/tests/test_create_mint.rs
```diff
@@ -0,0 +1,450 @@
...
+}
+
+#[test]
+fn test_create_mint_insufficient_funds() {
+    let program_id = spl_token_wrap::id();
+    let mut mollusk = Mollusk::new(&program_id, "spl_token_wrap");
+    mollusk_svm_programs_token::token2022::add_program(&mut mollusk);
+
+    let wrapped_mint_account = Pubkey::new_unique();
+    let wrapped_backpointer_account = Pubkey::new_unique();
```
joncinque, line 171: nit: same as above, proper derivation of addresses

program/tests/test_create_mint.rs
```diff
@@ -0,0 +1,450 @@
...
+    ];
+    mollusk.process_and_validate_instruction(
+        &instruction,
+        accounts,
+        &[Check::err(ProgramError::InsufficientFunds)],
+    );
+}
+
+#[test]
+fn test_create_mint_backpointer_insufficient_funds() {
```
joncinque, line 211: Along with these, can you add tests that improperly derived addresses fail?

program/tests/test_create_mint.rs
```diff
@@ -0,0 +1,450 @@
...
+        &program_id,
+        &wrapped_mint_address,
+        &wrapped_backpointer_address,
+        &unwrapped_mint_address,
+        &wrapped_token_program_id,
+        false,
+    );
+
+    let rent = &mollusk.sysvars.rent;
+    let unwrapped_mint_account = setup_token_2022_mint(rent);
```
joncinque, line 385: Can you add at least one extension to the token-2022 mint? The token-2022 tests should show a way to do that: https://github.com/solana-program/token-2022/blob/f69cc989c2a3f2657a6ad6041018468807223f08/program/src/extension/mod.rs#L1890

joncinque (commented):
Lots of great stuff here! Mostly small things, and I agree with all of Joe's points

program/src/processor.rs
buffalojoec: Gabe & I were discussing this a bit and I personally felt like it made more sense to pass the freeze authority through. It would stink to lose your freeze capabilities as a result of you wrapping your token.

joncinque: It's a different token, so I wouldn't assume anything carries over. For example, if I have freeze authority on a token, it gets wrapped, and I relinquish freeze authority on the unwrapped token, then I need to also know about the wrapped token to relinquish freeze authority there too. That seems strange.

Either way, this could be a great question for potential users!

program/src/instruction.rs
```diff
@@ -75,8 +78,148 @@ pub enum TokenWrapInstruction {
...
+
+/// Creates `CreateMint` instruction.
+pub fn create_mint(
+    program_id: &Pubkey,
+    wrapped_mint_address: &Pubkey,
+    wrapped_backpointer_address: &Pubkey,
+    unwrapped_mint_address: &Pubkey,
+    wrapped_token_program_id: &Pubkey,
+    idempotent: bool,
+) -> Instruction {
```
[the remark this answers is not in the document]
joncinque: Typically another function is fine -- there's the "raw" version of the instructor creator that takes everything, which is called by the "user-friendly" version that's easier to use.

program/src/instruction.rs
```diff
@@ -1,30 +1,33 @@
...
 /// Instructions supported by the Token Wrap program
-#[derive(Clone, Debug, PartialEq, TryFromPrimitive, IntoPrimitive)]
+#[derive(Clone, Debug, PartialEq)]
 #[repr(u8)]
 pub enum TokenWrapInstruction {
-    /// Create a wrapped token mint
+    /// Create a wrapped token mint. Assumes caller has pre-funded wrapped mint
+    /// and backpointer account. Supports both directions:
+    /// - spl-token to token-2022
+    /// - token-2022 to spl-token
```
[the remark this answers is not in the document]
joncinque: It's more interesting for token-2022 -- you could wrap an existing mint on token-2022 with new extensions! And in the future, as new protocols that adhere to SPL Token are developed, we may want to allow wrapping those too.

This program enables a neat extra use-case -- if someone wants to customize their own SPL Token program, they can do any wild behavior that they want, and then just wrap the tokens in SPL Token whenever needed

program/src/instruction.rs
```diff
@@ -1,30 +1,33 @@
 //! Program instructions
 
-use num_enum::{IntoPrimitive, TryFromPrimitive};
+use solana_program::instruction::{AccountMeta, Instruction};
+use solana_program::program_error::ProgramError;
+use solana_program::pubkey::Pubkey;
+use std::convert::TryInto;
```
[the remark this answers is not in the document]
joncinque: `solana-system-interface` will give you the system stuff, and `solana-sysvar` will give you the trait. There's a prerelease version available of `solana-sysvar` if you want, but the v2.2 branch should get cut very soon

program/tests/test_create_mint.rs
```diff
@@ -0,0 +1,773 @@
...
+        &wrapped_backpointer_address,
+        &unwrapped_mint_address,
+        &wrapped_token_program_id,
+        true,
+    );
+
+    // Simulating existing data on mint or backpointer
+    let mint_account_with_data = Account {
+        data: vec![1; 10],
+        owner: program_id,
```
joncinque, line 135: Per the previous comment, this needs to change too
```suggestion
        owner: wrapped_token_program_id,
```

program/tests/test_create_mint.rs
```diff
@@ -0,0 +1,773 @@
...
+        &incorrect_wrapped_mint_addr,
+        &wrapped_backpointer_address,
+        &unwrapped_mint_address,
+        &wrapped_token_program_id,
+        true,
+    );
+
+    let mint_account_with_data = Account {
+        data: vec![1; 10],
+        owner: program_id,
```
joncinque, line 181: Per the previous comment, this needs to change too
```suggestion
        owner: wrapped_token_program_id,
```

program/tests/test_create_mint.rs
```diff
@@ -0,0 +1,773 @@
...
+        &wrapped_mint_address,
+        &incorrect_wrapped_backpointer_addr,
+        &unwrapped_mint_address,
+        &wrapped_token_program_id,
+        true,
+    );
+
+    let mint_account_with_data = Account {
+        data: vec![1; 10],
+        owner: program_id,
```
joncinque, line 251: Per the previous comment, this needs to change too
```suggestion
        owner: wrapped_token_program_id,
```

program/tests/test_create_mint.rs
```diff
@@ -0,0 +1,773 @@
...
+        &wrapped_mint_address,
+        &wrapped_backpointer_address,
+        &unwrapped_mint_address,
+        &wrapped_token_program_id,
+        true,
+    );
+
+    let mint_account_with_data = Account {
+        data: vec![1; 10],
+        owner: program_id,
```
joncinque, line 288:
```suggestion
        owner: wrapped_token_program_id,
```

program/tests/test_create_mint.rs
```diff
@@ -426,18 +739,28 @@ fn test_successful_token_2022_to_spl_token() {
...
+            .unwrap()
+            .base;
+
     assert_eq!(wrapped_mint_data.decimals, MINT_DECIMALS);
     let expected_mint_authority = get_wrapped_mint_authority(&wrapped_mint_address);
     assert_eq!(
-        wrapped_mint_data.mint_authority.unwrap(),
+        wrapped_mint_data
+            .mint_authority
+            .ok_or(ProgramError::InvalidAccountData)
```
joncinque, line 752: nit: is this needed or is it possible to unwrap directly?

program/tests/test_create_mint.rs
```diff
@@ -426,18 +739,28 @@ fn test_successful_token_2022_to_spl_token() {
...
-    assert_eq!(wrapped_mint_data.supply, 0);
-    assert!(wrapped_mint_data.is_initialized);
+    assert_eq!(wrapped_mint_data.supply, PodU64::from(0));
+    assert_eq!(wrapped_mint_data.is_initialized, PodBool::from_bool(true));
     assert_eq!(
-        wrapped_mint_data.freeze_authority.unwrap(),
-        Pubkey::from_str_const(FREEZE_AUTHORITY)
+        wrapped_mint_data
+            .freeze_authority
+            .ok_or(ProgramError::InvalidAccountData)
```
joncinque, line 761: nit: same here, is this needed, or is it possible to just unwrap?

joncinque (commented):
Looking close! Mostly small things, then I'm happy to get this in

program/src/processor.rs
buffalojoec: > It's a different token, so I wouldn't assume anything carries over.

I disagree, actually. You're using a protocol to wrap one token in another, and the implication is that it's really just an SPL Token standard change. To introduce a change to your mint's configurations, which would be otherwise completely backwards compatible, seems unreasonable to me.

Furthermore, disabling the freeze authority on the wrapped token out of the gate is irreversible.

> I relinquish freeze authority on the unwrapped token, then I need to also know about the wrapped token to relinquish freeze authority there too. That seems strange.

This makes perfect sense to me.

program/src/instruction.rs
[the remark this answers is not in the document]
grod220: Nice. The system-interface was an easy import, however:

```
solana-sysvar = { version = "0.0.2", features = ["bincode"] }
```

Currently has build errors w/ rent/clock/slot... e.g:

```
error[E0277]: the trait bound `Rent: SysvarId` is not satisfied
   --> /Users/gabe/.cargo/registry/src/index.crates.io-6f17d22bba15001f/solana-sysvar-0.0.2/src/rent.rs:133:17
    |
133 | impl Sysvar for Rent {
    |                 ^^^^ the trait `SysvarId` is not implemented for `Rent`
    |
    = help: the following other types implement trait `SysvarId`:
              EpochRewards
              Fees
              Instructions
              RecentBlockhashes
              Rewards
              StakeHistory
note: required by a bound in `Sysvar`
   --> /Users/gabe/.cargo/registry/src/index.crates.io-6f17d22bba15001f/solana-sysvar-0.0.2/src/lib.rs:152:5
    |
151 | pub trait Sysvar:
    |           ------ required by a bound in this trait
152 |     SysvarId + Default + Sized + serde::Serialize + serde::de::DeserializeOwned
    |     ^^^^^^^^ required by this bound in `Sysvar`

```

Perhaps will resolve on release cut?

program/src/processor.rs
joncinque: Just to be clear, the whole concept for this program is that everything is permissionless -- anyone can create a wrapped mint, any unwrapped token holder can wrap tokens, any wrapped token holder can unwrap tokens.

The mint belongs entirely to the protocol, and not to the authorities on the unwrapped mint.

program/tests/test_create_mint.rs
grod220: Unfortunately, [COptionPod](https://github.com/solana-program/token-2022/blob/main/program/src/pod.rs#L162) does not expose a straight unwrap.

program/tests/test_create_mint.rs
grod220: Same as above with PodCOption

program/src/instruction.rs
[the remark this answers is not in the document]
joncinque: Ah right, we changed some stuff around for the sysvar id, so you have to enable the `sysvar` feature on the crate to get everything: https://github.com/anza-xyz/agave/blob/31cdd163d14061a20685da50cce0c079cd352f60/sdk/rent/Cargo.toml#L28

Let's just wait for v2.2

program/src/processor.rs
```diff
@@ -10,13 +10,14 @@ use {
...
     solana_msg::msg,
-    solana_program::{system_instruction, sysvar::Sysvar},
     solana_program_error::{ProgramError, ProgramResult},
     solana_program_pack::Pack,
     solana_pubkey::Pubkey,
     solana_rent::Rent,
+    solana_system_interface::instruction::{allocate, assign},
     spl_token_2022::{
         extension::PodStateWithExtensions, instruction::initialize_mint2, pod::PodMint,
+        solana_program::sysvar::Sysvar,
```
grod220, line 20: Given https://github.com/solana-program/token-wrap/pull/6#discussion_r1944507305 a work around is importing Sysvar from a token program. That was `solana-program` can be booted from cargo.toml. But perhaps that is a hacky workaround 🤷

program/tests/test_create_mint.rs
joncinque: Ah right, it only has the minimum required interface :sweat_smile: sorry!

program/src/lib.rs
```diff
@@ -7,9 +7,6 @@ pub mod instruction;
 pub mod processor;
 pub mod state;
 
-// Export current SDK types for downstream users building with a different SDK
-// version
-pub use solana_program;
```
grod220, line 12: Given the removal of `solana-program`, this does not feel relevant anymore (?)

program/src/instruction.rs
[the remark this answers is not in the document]
joncinque: to be totally clear, I mean "wait for v2.2 and keep using solana-program for now"

program/src/lib.rs
joncinque: Yeah feel free to remove

program/tests/test_create_mint.rs
```diff
@@ -163,42 +161,6 @@ fn test_idempotency_true_with_existing_invalid_accounts() {
         get_wrapped_mint_address(&unwrapped_mint_address, &wrapped_token_program_id);
     let wrapped_backpointer_address = get_wrapped_mint_backpointer_address(&wrapped_mint_address);
 
-    // Incorrectly derived wrapped mint address
-
-    let incorrect_wrapped_mint_addr = Pubkey::new_unique();
-
```
grod220: These test case removals were due to the mix of idempotency checks and PDA validations. Now that it is cleared up, the redundant test case in the idempotency test can be removed.

program/src/processor.rs
grod220: > Maybe it's better as an input parameter?

The issue with this is that the wrapped mint is shared between users. That means someone could call `CreateMint` and frontrun others with their preferred config.

>  I relinquish freeze authority on the unwrapped token, then I need to also know about the wrapped token to relinquish freeze authority there too. That seems strange.

If we take usdc for example, I think it would be a deal breaker if circle were unable to freeze wrapped tokens. If the wrapped version say would permeate a major protocol, in order for them to do so, they'd be forced to freeze the escrow account, essentially rugging all users of the wrapped token in that protocol. I'm inclined to think the freeze authority should carry over so the original entity retains the same per-account ability. In order to not block the remaining work on this program, perhaps we can move forward as is and revisit (?).

joncinque: Yeah definitely move forward! I still disagree, but that's not a problem.

I would stress that wrapped USDC is not Circle's token. If Circle decides to fork this program, then they create their special / official wrapped token, which can have whatever freeze authority / token extensions that they want. That special wrapped token, though, is not the same as the wrapped token created by this more general reference implementation.

But most importantly, we should get feedback from potential users.

program/src/processor.rs
joncinque: Haha very clever! To be honest, we'll likely remove the re-export from token-2022 with a breaking change once we update it to use the component crates, so best to just use `solana_program` directly

program/tests/test_create_mint.rs
joncinque: Makes sense!

joncinque (approved):
Looks good to me! But do get an approval from @buffalojoec too

program/src/processor.rs
```diff
@@ -1,25 +1,186 @@
...
+    let space = spl_token_2022::state::Mint::get_packed_len();
+
+    let rent = Rent::get()?;
+    let mint_rent_required = rent.minimum_balance(space);
+    if wrapped_mint_account.lamports() < mint_rent_required {
+        msg!(
+            "Error: wrapped_mint_account requires pre-funding of {} lamports",
+            mint_rent_required
+        );
+        return Err(ProgramError::InsufficientFunds);
```
buffalojoec, line 93: nit: you could use `ProgramError::AccountNotRentExempt` here if you want!

program/src/processor.rs
```diff
@@ -1,25 +1,186 @@
...
+    // Idempotency checks
+
+    if wrapped_mint_account.data_len() > 0 || wrapped_backpointer_account.data_len() > 0 {
+        msg!("Wrapped mint or backpointer account already initialized");
+        if !idempotent {
+            return Err(ProgramError::AccountAlreadyInitialized);
+        }
+        if wrapped_mint_account.owner != wrapped_token_program_account.key {
+            msg!("Wrapped mint account owner is not the expected token program");
+            return Err(ProgramError::InvalidAccountData);
```
buffalojoec, line 67:
```suggestion
            return Err(ProgramError::InvalidAccountOwner);
```
Also below this

buffalojoec (commented):
Awesome! This is super close, just a few tiny nits.

program/src/instruction.rs
```diff
@@ -48,10 +51,10 @@ pub enum TokenWrapInstruction {
     /// 8. `[signer]` Transfer authority on unwrapped token account
     /// 9. `..8+M` `[signer]` (Optional) M multisig signers on unwrapped token
     ///    account
-    ///
-    /// Data expected by this instruction:
-    ///   * little-endian `u64` representing the amount to wrap
-    Wrap,
+    Wrap {
```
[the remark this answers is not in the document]
grod220: Next PR covers this 👍

program/src/instruction.rs
```diff
@@ -75,8 +82,152 @@ pub enum TokenWrapInstruction {
...
+    multisig_signer_pubkeys: &[&Pubkey],
+    amount: u64,
+) -> Instruction {
+    let mut accounts = vec![
+        AccountMeta::new(*wrapped_token_account_address, false),
+        AccountMeta::new(*wrapped_mint_address, false),
+        AccountMeta::new(*unwrapped_escrow_address, false),
+        AccountMeta::new(*recipient_unwrapped_token_account_address, false),
+        AccountMeta::new_readonly(*unwrapped_mint_address, false),
+        AccountMeta::new_readonly(*wrapped_mint_authority_address, false),
```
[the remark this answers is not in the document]
grod220: This can actually be deleted and derived from within the instruction. Will update here, but the re-ordering and handling of this is in the next PR.

buffalojoec (approved):
Nice! Thanks for all of the back and forth!

PR #10, opened 2025-02-06
Wrap instruction

grod220:
After a Wrapped Mint is created, this instruction is used to escrow an unwrap token balance and receive the wrapped version to a recipient account. 

Features added:
- Process function for wrapping
- New builder pattern for tests (yay smaller test sizes!)
- Re-order Wrap accounts
- Escrow account derivation fns


program/src/instruction.rs
```diff
@@ -40,18 +40,20 @@ pub enum TokenWrapInstruction {
...
     ///    wrapped_token_program_id)`
-    /// 4. `[writeable]` Recipient wrapped token account
     /// 5. `[]` SPL Token program for unwrapped mint
     /// 6. `[]` SPL Token program for wrapped mint
-    /// 7. `[signer]` Transfer authority on unwrapped token account
-    /// 8. `..8+M` `[signer]` (Optional) M multisig signers on unwrapped token
+    /// 7. `[]` Unwrapped token mint
+    /// 8. `[]` Wrapped mint authority, address must be:
+    ///    `get_wrapped_mint_authority(wrapped_mint)`
+    /// 9. `..8+M` `[s]` (Optional) M multisig signers on unwrapped token
```
grod220, line 56: Re-ordered according to https://github.com/solana-program/token-wrap/pull/6#discussion_r1944951701

program/src/lib.rs
```diff
@@ -111,3 +111,37 @@ pub(crate) fn get_wrapped_mint_backpointer_address_with_seed(
...
+    authority: &Pubkey,
+    unwrapped_mint: &Pubkey,
+) -> (Pubkey, u8) {
+    Pubkey::find_program_address(&get_escrow_address_seeds(authority, unwrapped_mint), &id())
+}
+
+/// Derive the escrow address for a user's unwrapped tokens
+pub fn get_escrow_address(authority: &Pubkey, unwrapped_mint: &Pubkey) -> Pubkey {
+    get_escrow_address_with_seed(authority, unwrapped_mint).0
+}
```
grod220, line 147: Given an authority is used to derive the escrow address, this design essentially has an account per user/entity doing the wrapping. Another option is to not require that and bucket all those unwrapped tokens of the same mint together. If that is the direction, it should probably be done in `CreateMint`. Open for discussion on this.

program/src/processor.rs
```diff
@@ -164,6 +165,105 @@ pub fn process_create_mint(
...
+            return Err(ProgramError::InvalidAccountOwner);
+        }
+    }
+
+    // Transfer unwrapped tokens from user to escrow
+
+    let unwrapped_mint_data = unwrapped_mint.try_borrow_data()?;
+    let unwrapped_mint_state = PodStateWithExtensions::<PodMint>::unpack(&unwrapped_mint_data)?;
+    invoke_signed(
+        &spl_token_2022::instruction::transfer_checked(
```
grod220, line 223: For `mint_to` and `transfer_checked`, the functions needed to be derived from `spl_token_2022::instruction` (for both wrapping directions: spl_token<->token_2022). When using `spl_token::instruction`, it would throw.

program/tests/helpers/mint_builder.rs
grod220, line 1: A new builder pattern that made `CreateMint` re-use significantly easier. The tests become much more declarative this way as well.

program/src/processor.rs
```diff
@@ -164,6 +165,105 @@ pub fn process_create_mint(
...
+    let transfer_authority = next_account_info(account_info_iter)?;
+    let unwrapped_escrow = next_account_info(account_info_iter)?;
+    let unwrapped_token_account = next_account_info(account_info_iter)?;
+    let recipient_wrapped_token_account = next_account_info(account_info_iter)?;
+    let wrapped_mint = next_account_info(account_info_iter)?;
+    let unwrapped_token_program = next_account_info(account_info_iter)?;
+    let wrapped_token_program = next_account_info(account_info_iter)?;
+    let unwrapped_mint = next_account_info(account_info_iter)?;
+    let wrapped_mint_authority = next_account_info(account_info_iter)?;
+    let _signer_accounts = account_info_iter.as_slice();
```
grod220, line 186: Subsequent PR coming to handle multisig support.

program/tests/helpers/mint_builder.rs
buffalojoec: This builder pattern is pretty slick! I love how much code it allowed you to delete from the `CreateMint` tests. However, there's an arguably captious case to be made against scope-creep in program unit tests.

When you're testing `Instruction::CreateMint`, you don't want your tests to rely on other program instructions, since you're not testing those programs, you're testing _only_ `CreateMint`. So, we set up `Account` states manually, rather than invoking System or SPL Token(-2022).

Similarly, when you're testing `Instruction::Wrap` it's typically best not to obfuscate the unit tests with calls to `Instruction::CreateMint`. If any of the setup fails, it may take a contributor a bit of drilling to find out it's not `Instruction::Wrap` that has the bug, but `Instruction::CreateMint`.

All of that being said, there are some programs where contributors have done their setups similar to how you're doing it here, so I'll leave it up to you. This is just my opinion on program unit testing patterns!

buffalojoec (commented):
This looks great! I really only left comments on small stuff.

program/src/processor.rs
```diff
@@ -164,6 +165,105 @@ pub fn process_create_mint(
...
+
+    let expected_escrow = get_escrow_address(transfer_authority.key, unwrapped_mint.key);
+    if expected_escrow != *unwrapped_escrow.key {
+        msg!("Escrow address does not match the derived address");
+        return Err(ProgramError::InvalidAccountData);
+    }
+
+    {
+        let escrow_data = unwrapped_escrow.try_borrow_data()?;
+        let escrow_account = spl_token::state::Account::unpack(&escrow_data)?;
```
joncinque, line 211: Similar to the mint, we'll need `PodStateWithExtensions` in case the token account has any extensions, otherwise this will fail, ie: https://github.com/solana-program/token-2022/blob/a46685f457db8e5c68d7f43e1c032500b7cd258b/program/src/processor.rs#L1203

program/src/processor.rs
```diff
@@ -164,6 +165,105 @@ pub fn process_create_mint(
...
+            msg!("Unwrapped escrow token owner is not set to token-wrap program");
+            return Err(ProgramError::IncorrectAuthority);
+        }
+    }
+
+    // Transfer unwrapped tokens from user to escrow
+
+    let unwrapped_mint_data = unwrapped_mint.try_borrow_data()?;
+    let unwrapped_mint_state = PodStateWithExtensions::<PodMint>::unpack(&unwrapped_mint_data)?;
+    invoke_signed(
```
joncinque, line 222: Not for this PR, but it would be nice for this program to support transfer hooks, which means that we'll need the transfer hook helpers for transferring and pass in the right accounts.

There's an example program using the helper at https://github.com/solana-program/token-2022/blob/a46685f457db8e5c68d7f43e1c032500b7cd258b/clients/rust-legacy/transfer-hook-test-programs/swap/src/lib.rs#L32

More info about transfer-hooks at https://spl.solana.com/transfer-hook-interface

program/src/processor.rs
```diff
@@ -164,6 +165,105 @@ pub fn process_create_mint(
...
+            unwrapped_token_account.clone(),
+            unwrapped_mint.clone(),
+            unwrapped_escrow.clone(),
+            transfer_authority.clone(),
+        ],
+        &[],
+    )?;
+
+    // Mint wrapped tokens to recipient
+    let bump = get_wrapped_mint_authority_with_seed(wrapped_mint.key).1;
```
joncinque, line 243: This is calling `get_wrapped_mint_authority` and also `get_wrapped_mint_authority_with_seed`, which means that you'll be using the `find_program_address` syscall twice -- best to call `get_wrapped_mint_authority_with_seed` earlier and reusing the bump here

program/tests/helpers/mint_builder.rs
```diff
@@ -0,0 +1,226 @@
...
+};
+
+pub struct CreateMintResult {
+    pub unwrapped_mint: KeyedAccount,
+    pub wrapped_mint: KeyedAccount,
+    pub wrapped_backpointer: KeyedAccount,
+}
+
+#[derive(Debug, Clone)]
+#[cfg(test)]
```
joncinque, line 19: nit: why does this have `cfg(test)`? It's all testing code, right? We might just need to add an `#[allow(dead_code)]` somewhere, potentially on the whole file

program/tests/helpers/mint_builder.rs
```diff
@@ -0,0 +1,226 @@
...
+    #[cfg(test)]
+    pub fn pair(&self) -> (Pubkey, Account) {
+        (self.key, self.account.clone())
+    }
+}
+
+#[derive(Debug, Clone, Copy)]
+pub enum TokenProgram {
+    SplToken,
+    Token2022,
```
joncinque, line 36: nit: token-2022 is also an SPL program, so let's be consistent and add the `Spl*` prefix to both, or neither

program/tests/helpers/mint_builder.rs
```diff
@@ -0,0 +1,226 @@
...
+
+    pub fn keyed_account(&self) -> (Pubkey, Account) {
+        match self {
+            TokenProgram::SplToken => mollusk_svm_programs_token::token::keyed_account(),
+            TokenProgram::Token2022 => mollusk_svm_programs_token::token2022::keyed_account(),
+        }
+    }
+}
+
+pub struct MintBuilder<'a> {
```
joncinque, line 55: nit: since this is wrapping `CreateMint`, should it be `CreateMintBuilder`?

program/tests/helpers/mint_builder.rs
```diff
@@ -0,0 +1,226 @@
...
+    idempotent: bool,
+    checks: Vec<Check<'a>>,
+}
+
+impl<'a> Default for MintBuilder<'a> {
+    fn default() -> Self {
+        Self {
+            mollusk: init_mollusk(),
+            wrapped_token_program: TokenProgram::Token2022,
+            wrapped_token_program_addr: None,
```
joncinque, line 75: Is this needed? `wrapped_token_program.id()` should give the info, no?

program/tests/helpers/common.rs
```diff
@@ -0,0 +1,79 @@
...
+}
+
+pub const MINT_DECIMALS: u8 = 12;
+pub const MINT_SUPPLY: u64 = 500_000_000;
+pub const FREEZE_AUTHORITY: Pubkey =
+    Pubkey::from_str_const("11111115q4EpJaTXAZWpCg3J2zppWGSZ46KXozzo9");
+
+fn token_2022_with_extension_data(supply: u64) -> Vec<u8> {
+    let mint_size =
+        ExtensionType::try_calculate_account_len::<PodMint>(&[ExtensionType::MintCloseAuthority])
```
joncinque, line 37: To test the behavior when the token account has an extension, you can add the `TransferFeeConfig` extension to the mint

program/tests/test_create_mint.rs
joncinque, line 1: In general, it's better to do test refactors in a separate PR.

program/tests/test_create_mint.rs
```diff
@@ -1,256 +1,103 @@
...
-    mollusk.process_and_validate_instruction(
-        &instruction,
-        accounts,
-        &[Check::err(ProgramError::AccountAlreadyInitialized)],
-    );
+    // Test case 1: mint already exists
+    MintBuilder::default()
+        .wrapped_mint_account(account_with_data.clone())
+        .check(Check::err(ProgramError::AccountAlreadyInitialized))
+        .execute();
```
joncinque, line 31: Wow, I never understood why people liked builder patterns so much, but this looks really nice!

program/tests/test_create_mint.rs
```diff
@@ -1,256 +1,103 @@
...
-        &wrapped_backpointer_address,
-        &unwrapped_mint_address,
-        &wrapped_token_program_id,
-        true,
-    );
-
     let mint_account_with_data = Account {
         data: vec![1; 10],
-        owner: wrapped_token_program_id,
+        owner: spl_token_wrap::id(),
```
joncinque, line 81: The comment only mentions that the backpointer account has an incorrect owner, but this has a wrong owner on the mint too -- since the incorrect owner on the mint was tested earlier, should this be correct?

joncinque (commented):
Looking good! Mostly small things

program/src/processor.rs
```diff
@@ -164,6 +165,105 @@ pub fn process_create_mint(
...
+    let wrapped_mint_authority = next_account_info(account_info_iter)?;
+    let _signer_accounts = account_info_iter.as_slice();
+
+    // Validate accounts
+
+    let expected_wrapped_mint =
+        get_wrapped_mint_address(unwrapped_mint.key, wrapped_token_program.key);
+    if expected_wrapped_mint != *wrapped_mint.key {
+        msg!("Wrapped mint address does not match the derived address");
+        return Err(ProgramError::InvalidAccountData);
```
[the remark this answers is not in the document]
buffalojoec: I think that's still a little bit vague, but it's definitely better.

Here's an example of custom errors. Basically they all pipe to `ProgramError::Custom(u32)`.
https://github.com/solana-program/token-2022/blob/main/program/src/error.rs

program/tests/helpers/mint_builder.rs
grod220: Ah, a vestige of having `dead_code` linting issues and debugging. Will remove. Turns out I just needed to do this in `/tests/helpers/mod.rs`

```diff
- mod common;
+ pub mod common;
```

program/tests/helpers/mint_builder.rs
grod220: Ah, I was wondering the right syntax for this. Will fix.

program/tests/helpers/mint_builder.rs
grod220: If none is passed, it will indeed grab from there. It was added as a test param [for this test](https://github.com/solana-program/token-wrap/pull/10/files#diff-50e76d72138fd69729c6e469dc9201b57b77542f9c631bbe8f5077821dd473b3R159-R163) that tests if someone passes the wrong address for the token program. But maybe not needed?

program/tests/test_create_mint.rs
grod220: Yes, that's fair. In the original iteration, the MintBuilder was a dependency on the WrapBuilder. But after implementing [this feedback](https://github.com/solana-program/token-wrap/pull/10#discussion_r1948846225) it's no longer the case. So it does feel pretty independent now.

program/tests/test_create_mint.rs
grod220: Ah, a test bug! This is a good argument for custom errors to be used to disambiguate as `ProgramError::InvalidAccountOwner` is used in both cases. This passes because it just repeats case with the wrong owner on the mint. Will fix!

program/src/processor.rs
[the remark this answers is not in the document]
grod220: Oh I like this a lot, will implement 👍

program/src/error.rs
```diff
@@ -0,0 +1,78 @@
...
+    #[error("Unwrapped escrow token owner is not set to token-wrap program")]
+    EscrowOwnerMismatch,
+
+    // 5
+    /// Wrapped mint account owner is not the expected token program
+    #[error("Wrapped mint account owner is not the expected token program")]
+    InvalidWrappedMintOwner,
+    /// Wrapped backpointer account owner is not the expected token wrap program
+    #[error("Wrapped backpointer account owner is not the expected token wrap program")]
+    InvalidBackpointerOwner,
```
buffalojoec, line 40: Nice! I'm all for explicit errors.

program/tests/helpers/mint_builder.rs
joncinque: Ah sorry, I forgot to remove this comment after going through the tests -- yep looks good!

program/src/error.rs
```diff
@@ -0,0 +1,78 @@
...
+    BackpointerMismatch,
+    /// Wrap amount should be positive
+    #[error("Wrap amount should be positive")]
+    ZeroWrapAmount,
+    /// Wrapped mint authority does not match expected PDA
+    #[error("Wrapped mint authority does not match expected PDA")]
+    MintAuthorityMismatch,
+    /// Unwrapped escrow token owner is not set to token-wrap program
+    #[error("Unwrapped escrow token owner is not set to token-wrap program")]
+    EscrowOwnerMismatch,
```
joncinque, line 32: nit: the escrow owner is a PDA, and more specifically, the wrapped mint authority, so probably clearer to say `Unwrapped escrow token owner is not set to expected PDA`

program/src/lib.rs
```diff
@@ -111,3 +112,37 @@ pub(crate) fn get_wrapped_mint_backpointer_address_with_seed(
...
+pub(crate) fn get_escrow_address_with_seed(
+    authority: &Pubkey,
+    unwrapped_mint: &Pubkey,
+) -> (Pubkey, u8) {
+    Pubkey::find_program_address(&get_escrow_address_seeds(authority, unwrapped_mint), &id())
+}
+
+/// Derive the escrow address for a user's unwrapped tokens
+pub fn get_escrow_address(authority: &Pubkey, unwrapped_mint: &Pubkey) -> Pubkey {
+    get_escrow_address_with_seed(authority, unwrapped_mint).0
```
joncinque, line 147: Can probably remove all this, right?

program/tests/helpers/mint_builder.rs
joncinque, line 1: nit: rename this to `create_mint_builder.rs` to be totally clear?

program/tests/helpers/wrap_builder.rs
```diff
@@ -0,0 +1,299 @@
...
+        let wrapped_mint_addr = get_wrapped_mint_address(&unwrapped_mint_addr, &token_program.id());
+        let mint_authority = get_wrapped_mint_authority(&wrapped_mint_addr);
+
+        self.wrapped_mint.clone().unwrap_or(KeyedAccount {
+            key: wrapped_mint_addr,
+            account: setup_mint(token_program, &self.mollusk.sysvars.rent, mint_authority),
+        })
+    }
+
+    pub fn setup_token_account(&self, wrapped_mint: &KeyedAccount) -> KeyedAccount {
```
joncinque, line 121: Although the mint on token-2022 is correctly configured with the `TransferFeeConfig`, the token accounts need to also be configured with the `TransferFeeAmount` extension, otherwise the token accounts won't work properly. Also, we won't get to test the behavior of when a token account has an extension

joncinque (commented):
Just a few last bits, then this is good to go on my side, along with Joe's comments.

In the future, please do things like error refactors in another PR, since that would also make the review simpler.

program/tests/helpers/wrap_builder.rs
grod220: Ah I see. Updated this with proper extension initialization.

program/src/error.rs
```diff
@@ -0,0 +1,78 @@
...
+impl TokenWrapError {
+    /// Log the error and convert it into a `ProgramError`.
+    ///
+    /// Call this method in place of a standard conversion when you want to have
+    /// the custom error logged.
+    pub fn log_into(self) -> ProgramError {
+        self.print::<TokenWrapError>();
+        ProgramError::Custom(self as u32)
+    }
+}
```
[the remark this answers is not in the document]
buffalojoec: > Should we convert all error returns to the `?` propagation style?

Up to you! I like the simpler one, but agreed they should all match. Either one works to convert the error type using `From`.

```rust
        if !idempotent {
            Err(ProgramError::AccountAlreadyInitialized)?
        }
        if wrapped_mint_account.owner != wrapped_token_program_account.key {
            Err(TokenWrapError::InvalidWrappedMintOwner)?
        }
```

program/src/error.rs
```diff
@@ -0,0 +1,78 @@
...
+    fn print<E>(&self)
+    where
+        E: 'static + Error + DecodeError<E> + PrintProgramError + num_traits::FromPrimitive,
+    {
+        msg!(
+            "{error_type}: {variant:?} - {description}",
+            error_type = E::type_of(),
+            variant = self,
+            description = self
+        );
```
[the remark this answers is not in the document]
buffalojoec: Yeah, it doesn't show the variant name, but the variant's main purpose is to provide a code in the program's return. You'll see on failure it will say `Custom program error 0x1`, etc.

I would lean toward just the simple `to_string` since it's really just about logging, and the code is returned and can be decoded by the client.

buffalojoec (commented):
This looks good to me, just want to reach consensus on the error comments and I'll approve!

buffalojoec (approved):
Lgtm! Make sure to get a stamp from @joncinque as well!

joncinque (approved):
Looks great to me, thanks for addressing everything!

PR #11, opened 2025-02-10
`Wrap` multisig support

grod220:
A smaller follow up on https://github.com/solana-program/token-wrap/pull/10. Enables multisig support for Wrap instruction.


program/tests/test_wrap.rs
```diff
@@ -83,10 +85,16 @@ fn assert_wrap_result(starting_amount: u64, wrap_amount: u64, wrap_result: &Wrap
...
-    //     spl_token_2022::state::Mint::unpack(&wrap_result.wrapped_mint.
-    // account.data).unwrap(); assert_eq!(wrapped_mint.supply, wrap_amount);
+    // Verify wrapped mint supply increased
+    if wrap_result.wrapped_mint.account.owner == spl_token::id() {
+        let mint = spl_token::state::Mint::unpack(&wrap_result.wrapped_mint.account.data).unwrap();
+        assert_eq!(mint.supply, MINT_SUPPLY + wrap_amount);
+    } else {
+        let mint =
+            PodStateWithExtensions::<PodMint>::unpack(&wrap_result.wrapped_mint.account.data)
+                .unwrap();
```
joncinque, line 95: This works for any mint on token or token-2022, so feel free to use it without checking the owner!

program/src/processor.rs
```diff
@@ -219,24 +219,34 @@ pub fn process_wrap(_program_id: &Pubkey, accounts: &[AccountInfo], amount: u64)
...
+
+    let transfer_accounts = vec![
+        unwrapped_token_account.clone(),
+        unwrapped_mint.clone(),
+        unwrapped_escrow.clone(),
+        transfer_authority.clone(),
+    ]
+    .into_iter()
+    .chain(multisig_signer_accounts.iter().cloned())
+    .collect::<Vec<_>>();
```
joncinque, line 236: This might not need to be addressed, but all of the allocations could be costly -- just out of curiosity, how much does this code increase CU usage for non multisig cases?

joncinque (commented):
Just a tiny thing and a question, looks great otherwise!

program/src/processor.rs
buffalojoec: Is it possible to just access the `&[AccountInfo]` with references, and then pass a subslice of that into the CPI call? If we matched things up with the transfer accounts ordering, that is.

grod220: > how much does this code increase CU usage for non multisig cases?

- Wrap code without multisig updates:
  -  `test_successful_spl_token_wrap` -> 20298 compute units
- This PR's updates:
  - `test_successful_spl_token_wrap` -> 19230 compute units (strangely lower)
  - `test_wrap_with_spl_token_multisig` -> 29580 compute units
  - `test_wrap_with_token_2022_multisig` -> 25088 compute units

> Is it possible to just access the &[AccountInfo] with references, and then pass a subslice of that into the CPI call?

We can't concatenate two slices together without a Vec. But pre-allocating the Vec may be slightly better:

```rust
    let mut transfer_accounts = Vec::with_capacity(4 + multisig_signer_accounts.len());
    transfer_accounts.push(unwrapped_token_account.clone());
    transfer_accounts.push(unwrapped_mint.clone());
    transfer_accounts.push(unwrapped_escrow.clone());
    transfer_accounts.push(transfer_authority.clone());
    transfer_accounts.extend_from_slice(multisig_signer_accounts);
```

- `test_wrap_with_spl_token_multisig` -> 29297 compute units
- `test_wrap_with_token_2022_multisig` -> 24805 compute units

It looks like a tiny impact, but I don't really have a sense of how much that is. Worthwhile?

joncinque: Pre-allocating makes sense to me! As long as it gives a benefit without being too hard to read, it's worth it to me.

Until we come up with a better API for CPIs, this might be the best we can do, outside of Joe's suggestion of lining up the accounts.

grod220: @buffalojoec ah I see what you mean. Aligning account order does indeed work:
- `test_wrap_with_spl_token_multisig` -> 28586 compute units (-711)
- `test_wrap_with_token_2022_multisig` -> 24094 compute units (-711)

But the instruction API now reads like so. Worth it?
```rust
    /// 0. `[w]` Recipient wrapped token account
    /// 1. `[w]` Wrapped mint, must be initialized, address must be:
    ///    `get_wrapped_mint_address(unwrapped_mint_address,
    ///    wrapped_token_program_id)`
    /// 2. `[]` Wrapped mint authority, address must be:
    ///    `get_wrapped_mint_authority(wrapped_mint)`
    /// 3. `[]` SPL Token program for unwrapped mint
    /// 4. `[]` SPL Token program for wrapped mint
    /// 5. `[w]` Unwrapped token account to wrap
    ///    `get_wrapped_mint_authority(wrapped_mint_address)`
    /// 6. `[]` Unwrapped token mint
    /// 7. `[w]` Escrow of unwrapped tokens, must be owned by:
    ///    `get_wrapped_mint_authority(wrapped_mint_address)`
    /// 8. `[s]` Transfer authority on unwrapped token account. Not required to
    ///    be a signer if it's a multisig.
    /// 9. `..8+M` `[s]` (Optional) M multisig signers on unwrapped token
    ///    account.
    Wrap {
        /// little-endian `u64` representing the amount to wrap
        amount: u64,
    },
```

joncinque (approved):
Looks great!

buffalojoec (approved):
Ship it!

PR #13, opened 2025-02-14
Unwrap instruction

grod220:
The remaining instruction to complete a full user-flow with wrapping:
```
CreateMint -> Wrap -> Unwrap (this PR)
```

The unwrap action burns the wrapped tokens and transfers the equivalent amount from the escrow account to the recipient.

### DIFF
--- a/program/src/instruction.rs
+++ b/program/src/instruction.rs
@@ -69,18 +69,20 @@ pub enum TokenWrapInstruction {
     ///
     /// Accounts expected by this instruction:
     ///
-    /// 0. `[writeable]` Wrapped token account to unwrap
-    /// 1. `[writeable]` Wrapped mint, address must be:
+    /// 0. `[writeable]` Escrow of unwrapped tokens, must be owned by:
+    ///    `get_wrapped_mint_authority(wrapped_mint_address)`
+    /// 1. `[writeable]` Recipient unwrapped tokens
+    /// 2. `[]` Wrapped mint authority, address must be:
+    ///    `get_wrapped_mint_authority(wrapped_mint)`
+    /// 3. `[]` Unwrapped token mint
+    /// 4. `[]` SPL Token program for wrapped mint
+    /// 5. `[]` SPL Token program for unwrapped mint
+    /// 6. `[writeable]` Wrapped token account to unwrap
+    /// 7. `[writeable]` Wrapped mint, address must be:
     ///    `get_wrapped_mint_address(unwrapped_mint_address,
     ///    wrapped_token_program_id)`
-    /// 2. `[writeable]` Escrow of unwrapped tokens, must be owned by:
-    ///    `get_wrapped_mint_authority(wrapped_mint_address)`
-    /// 3. `[writeable]` Recipient unwrapped tokens
-    /// 4. `[]` Unwrapped token mint
-    /// 5. `[]` SPL Token program for wrapped mint
-    /// 6. `[]` SPL Token program for unwrapped mint
-    /// 7. `[signer]` Transfer authority on wrapped token account
-    /// 8. `..8+M` `[signer]` (Optional) M multisig signers on wrapped token
+    /// 8. `[signer]` Transfer authority on wrapped token account
+    /// 9. `..8+M` `[signer]` (Optional) M multisig signers on wrapped token
     ///    account
     Unwrap {
         /// little-endian `u64` representing the amount to unwrap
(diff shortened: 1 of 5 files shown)


program/src/instruction.rs
```diff
@@ -69,18 +69,20 @@ pub enum TokenWrapInstruction {
...
-    /// 2. `[writeable]` Escrow of unwrapped tokens, must be owned by:
-    ///    `get_wrapped_mint_authority(wrapped_mint_address)`
-    /// 3. `[writeable]` Recipient unwrapped tokens
-    /// 4. `[]` Unwrapped token mint
-    /// 5. `[]` SPL Token program for wrapped mint
-    /// 6. `[]` SPL Token program for unwrapped mint
-    /// 7. `[signer]` Transfer authority on wrapped token account
-    /// 8. `..8+M` `[signer]` (Optional) M multisig signers on wrapped token
+    /// 8. `[signer]` Transfer authority on wrapped token account
+    /// 9. `..8+M` `[signer]` (Optional) M multisig signers on wrapped token
```
grod220, line 85: Re-arranged in order so that accounts can be sub-sliced intothe transfer_checked account slice

joncinque (approved):
This can go in as is for me, just the comment about more checks, but I don't really think they're needed.

Please wait for Joe's feedback too though

program/src/processor.rs
```diff
@@ -251,6 +251,95 @@ pub fn process_wrap(_program_id: &Pubkey, accounts: &[AccountInfo], amount: u64)
...
+    let wrapped_mint_authority = next_account_info(account_info_iter)?;
+    let unwrapped_mint = next_account_info(account_info_iter)?;
+    let wrapped_token_program = next_account_info(account_info_iter)?;
+    let unwrapped_token_program = next_account_info(account_info_iter)?;
+    let wrapped_token_account = next_account_info(account_info_iter)?;
+    let wrapped_mint = next_account_info(account_info_iter)?;
+    let transfer_authority = next_account_info(account_info_iter)?;
+    let multisig_signer_accounts = account_info_iter.as_slice();
+
+    // Validate accounts
```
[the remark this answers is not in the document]
buffalojoec: I would say no, we don't need account state and owner checks, since everything has to go through an SPL-Token instruction (`mint_to`, `burn`, `transfer_checked`) and the respective token program should have those checks. It's better to leave those checks up to the token program anyway, since this program could theoretically be used with alternate token programs.

Verifying that you're unwrapping to the proper token (and wrapping in the previous instruction) is done by the PDA derivations, so I think we're covered.

program/src/processor.rs
```diff
@@ -251,6 +251,95 @@ pub fn process_wrap(_program_id: &Pubkey, accounts: &[AccountInfo], amount: u64)
...
+        Err(TokenWrapError::MintAuthorityMismatch)?
+    }
+
+    {
+        let escrow_data = unwrapped_escrow.try_borrow_data()?;
+        let escrow_account = PodStateWithExtensions::<PodAccount>::unpack(&escrow_data)?;
+        if escrow_account.base.owner != expected_authority {
+            Err(TokenWrapError::EscrowOwnerMismatch)?
+        }
+    }
```
buffalojoec, line 292: This is another one of those checks that could be left up to the token program, which would also check the `base.mint`.

program/tests/test_unwrap.rs
```diff
@@ -0,0 +1,236 @@
...
+}
+
+#[test]
+fn test_unwrap_amount_exceeds_unwrappers_balance() {
+    let wrapped_balance = 1_000;
+    let unwrap_amount = 42_000;
+
+    UnwrapBuilder::default()
+        .unwrapped_token_program(TokenProgram::SplToken)
+        .wrapped_token_program(TokenProgram::SplToken2022)
```
buffalojoec, line 67: Are these needed for this test?

program/tests/test_unwrap.rs
```diff
@@ -0,0 +1,236 @@
...
+    escrow_starting_amount: u64,
+    unwrap_amount: u64,
+    unwrap_result: &UnwrapResult,
+) {
+    // Verify wrapped tokens were burned (source account)
+    let wrapped_token = PodStateWithExtensions::<PodAccount>::unpack(
+        &unwrap_result.wrapped_token_account.account.data,
+    )
+    .unwrap();
+    assert_eq!(wrapped_token.base.amount, 0.into());
```
buffalojoec, line 85: Shouldn't this be `source_starting_amount - unwrap_amount`?

buffalojoec (approved):
Lgtm, just little things. Approving!

program/tests/test_unwrap.rs
grod220: Can toss these 👍

program/tests/test_unwrap.rs
grod220: Ah, indeed that would be more accurate. At the moment, it's assuming the source starting amount is the same amount as what you want to unwrap. Will update!

PR #15, opened 2025-02-27
Transfer hook support

grod220:
If the mint has support for the TransferHooks extension, the program will now execute them upon a wrap or unwrap action.

Makes use of helpers exposed by token-2022 repo. However, there is a dependency on https://github.com/solana-program/token-2022/pull/233 for multisig support.

New `test-transfer-hook` program created to test out new functionality.


package.json
```diff
@@ -10,7 +10,7 @@
     "template:upgrade": "zx ./scripts/upgrade-template.mjs",
     "rust:spellcheck": "cargo spellcheck --code 1",
     "rust:audit": "zx ./scripts/rust/audit.mjs",
-    "rust:semver": "cargo semver-checks"
+    "rust:semver": "cargo semver-checks --manifest-path program/Cargo.toml"
```
grod220, line 13: Avoids issues with `test-transfer-hook` from being checked

scripts/rust/test-sbf.mjs
```diff
@@ -1,11 +1,13 @@
...
 } from '../utils.mjs';
 
 const [folder, ...args] = cliArguments();
-const manifestPath = path.join(workingDirectory, folder, 'Cargo.toml');
+const mainProgramManifestPath = path.join(workingDirectory, folder, 'Cargo.toml');
+const testProgramManifestPath = path.join(workingDirectory, folder, 'tests', 'helpers', 'test-transfer-hook', 'Cargo.toml');
 
-await $`RUST_LOG=error cargo test-sbf --manifest-path ${manifestPath} ${args}`;
+await $`RUST_LOG=error cargo test-sbf --manifest-path ${testProgramManifestPath} ${args}`;
+await $`RUST_LOG=error cargo test-sbf --manifest-path ${mainProgramManifestPath} ${args}`;
```
grod220, line 13: Important to add the test program here as the main program tests depend on the .so file

program/src/instruction.rs
```diff
@@ -172,6 +172,7 @@ pub fn wrap(
     unwrapped_escrow_address: &Pubkey,
     transfer_authority_address: &Pubkey,
     multisig_signer_pubkeys: &[&Pubkey],
+    transfer_hook_metas: Vec<AccountMeta>,
```
buffalojoec, line 175: nit: This can be a slice.

buffalojoec: Also, if you want, you can leave these helpers alone as the "base" helpers, and you can roll a set of additional helpers that look just like [these ones in Token-2022](https://github.com/solana-program/token-2022/blob/4829c03a0b34d86db3b663262883a434268874fa/program/src/offchain.rs#L37). You could even use an `offchain` module to keep with the pattern in the interface and the Token-2022 program.

program/src/processor.rs
```diff
@@ -11,15 +11,16 @@ use {
     solana_account_info::{next_account_info, AccountInfo},
     solana_cpi::{invoke, invoke_signed},
     solana_msg::msg,
-    solana_program::sysvar::Sysvar,
     solana_program_error::{ProgramError, ProgramResult},
     solana_program_pack::Pack,
     solana_pubkey::Pubkey,
     solana_rent::Rent,
     solana_system_interface::instruction::{allocate, assign},
+    spl_token::solana_program::sysvar::Sysvar,
```
buffalojoec, line 19: Is this used?

program/src/processor.rs
```diff
@@ -283,23 +275,39 @@ pub fn process_unwrap(accounts: &[AccountInfo], amount: u64) -> ProgramResult {
...
+    let mut multisig_signers = Vec::new();
+    let mut transfer_hook_accounts = Vec::new();
+    for account in additional_accounts {
+        if account.is_signer {
+            multisig_signers.push(account);
+        } else {
+            transfer_hook_accounts.push(account.clone());
+        }
+    }
+
```
buffalojoec, line 289: A transfer hook account extra can also be a signer. What happens if you just pass all the accounts to `burn` and then use the onchain transfer helper for `transfer`?

grod220: So this works for burn (will update this), however it's tricker for `transfer`. If I use the helper `invoke_transfer_checked()` and signed multisig accounts were passed to the `unwrap` instruction (relevant to the burn), it will attempt to incorrectly add those to the escrow transfer (it should be PDA authorized transfer) https://github.com/solana-program/token-2022/blob/4829c03a0b34d86db3b663262883a434268874fa/program/src/instruction.rs#L1689-L1692 and throw `MissingRequiredSignature`.

Sounds like we have a few options:
- Not support signed extra transfer hook accounts (feels less than ideal)
- Update `invoke_transfer_checked()` helper in token-2022 repo to properly detect multisig accounts and only add those to the transfer invocation. Aka, instead of this https://github.com/solana-program/token-2022/blob/4829c03a0b34d86db3b663262883a434268874fa/program/src/onchain.rs#L49-L58, parse specifically multisig accounts like so: https://github.com/solana-program/token-2022/blob/4829c03a0b34d86db3b663262883a434268874fa/program/src/processor.rs#L1949-L1956

Thoughts?

program/Cargo.toml
```diff
@@ -15,29 +15,33 @@ test-sbf = []
...
 [dev-dependencies]
-mollusk-svm = "0.0.15"
-mollusk-svm-programs-token = "0.0.15"
-solana-account = "2.1.13"
+mollusk-svm = "0.1.1"
+mollusk-svm-programs-token = "0.1.1"
+spl-tlv-account-resolution = "0.9.0"
+solana-account = "2.2.1"
+solana-system-program = "2.2.0"
+test-transfer-hook = { path = "tests/helpers/test-transfer-hook", features = ["no-entrypoint"] }
```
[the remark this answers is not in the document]
grod220: Originally I was following the pattern that Mollusk had for its [CPI tests](https://github.com/anza-xyz/mollusk/tree/main/test-programs). I worry if we don't do it this way, it doesn't quite emulate a program-to-program test. Perhaps I am off though?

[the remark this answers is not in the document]
buffalojoec: It should be the same, but I suppose there's no harm in testing with _true_ BPF-BPF cross-program invocations, as you're doing here. Maybe we can leave as you have it, then.

One small nit would be to consider committing the program ELF for the test program, like we do in Token-2022, if you think it makes life easier for CI/etc.
https://github.com/solana-program/token-2022/tree/main/clients/rust-legacy/tests/fixtures

program/src/instruction.rs
buffalojoec: > Given the accounts require owned `AccountMeta`, if this were a slice it would need to clone:

Ah, gotcha. If `AccountMeta` can't be dereferenced, then yeah makes sense.

program/src/processor.rs
buffalojoec: Where?

program/src/processor.rs
buffalojoec: Even if we use the helpers it will add those unused accounts?

The transfer hook helpers are supposed to pick out each "extra meta" individually. So, any multisig signers not relevant for the transfer should never make it in there.

Maybe here, since we are burning and not transferring, we know we don't want to add multisig signers to the `transfer` instruction (only the `burn`). So, we can probably circumvent `invoke_transfer_checked` and just build things out with `spl_transfer_hook_interface::onchain::add_extra_accounts_for_execute_cpi` directly.

What do you think?

grod220: Actually, was a bit confused. There are two problems. Both require multisig account detection.

1. `invoke_transfer_checked()` does not support signed extra transfer hook accounts. 
   - Going to open a subsequent PR to [this PR](https://github.com/solana-program/token-2022/pull/233) to address that. There is a current bug where it adds a redundant AccountMeta to the instruction likely messing up account ordering. 
2. The Unwrap call to `invoke_transfer_checked()` includes possible multisig accounts which aren't relevant to the PDA escrow transfer (creates a transfer authority non-signature account meta which causes a `MissingRequiredSignature` error in token-2022 program).
   - The challenge with using `add_extra_accounts_for_execute_cpi()` directly is that I would still need to call it conditionally (dependent on mint hook), assemble the instruction & remove the multisig accounts from the accounts list. If I can just do the removal step, I can re-use `invoke_transfer_checked()` all the same.

program/Cargo.toml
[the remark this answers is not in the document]
grod220: My gut feeling is that committing the .so file would add complexity. Or maybe I don't see the part of the code it would allow us to exclude. At the moment, when you build, it grabs the current test crate code and tests against that. If I were to commit the .so and point to that, I'd need to setup additional CI/CD steps to ensure those are updated with a `git diff` against the live code version.

program/src/processor.rs
grod220: https://github.com/solana-program/token-wrap/blob/main/program/src/processor.rs#L85

program/src/processor.rs
grod220: Alrighty, [this PR](https://github.com/solana-program/token-2022/pull/253) should do the trick. It both fixes the transfer hook bug & exposes a helper to extract the multisig keys to avoid problem #2.

program/src/processor.rs
buffalojoec: Ohh I see, I missed where the original was on the diff. Why did it need to change to the `spl_token` reexport? I'm assuming because of the version bumps?

You can just use the `solana-sysvar` crate, but again, this is a good case for making versions bumps separate PRs.

program/Cargo.toml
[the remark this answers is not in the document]
buffalojoec: Ok sure, if you anticipate the counter program will change in the future, we can keep it as-is.

program/src/instruction.rs
```diff
@@ -137,99 +132,3 @@ impl TokenWrapInstruction {
...
-    let accounts = vec![
-        AccountMeta::new(*wrapped_mint_address, false),
-        AccountMeta::new(*wrapped_backpointer_address, false),
-        AccountMeta::new_readonly(*unwrapped_mint_address, false),
-        AccountMeta::new_readonly(solana_system_interface::program::id(), false),
-        AccountMeta::new_readonly(*wrapped_token_program_id, false),
-    ];
-    let data = TokenWrapInstruction::CreateMint { idempotent }.pack();
-    Instruction::new_with_bytes(*program_id, &data, accounts)
-}
```
buffalojoec, line 159: Sorry, I missed your [last line in your comment here](https://github.com/solana-program/token-wrap/pull/15#discussion_r1975579790), but I don't think these belong in an `offchain` module. We always define these sort of functions that return `Instruction` in the `instruction.rs` module.

What I actually meant was you could leave these three functions alone, as they were before this PR, and then have additional helpers in `offchain.rs` - such as `create_wrap_instruction_with_extra_metas` - that will literally just call `spl_token_2022::offchain::add_extra_account_metas` like the ones I linked do.

program/src/offchain.rs
```diff
@@ -0,0 +1,113 @@
...
+    wrapped_mint_address: &Pubkey,
+    wrapped_mint_authority_address: &Pubkey,
+    unwrapped_token_program_id: &Pubkey,
+    wrapped_token_program_id: &Pubkey,
+    unwrapped_token_account_address: &Pubkey,
+    unwrapped_mint_address: &Pubkey,
+    unwrapped_escrow_address: &Pubkey,
+    transfer_authority_address: &Pubkey,
+    multisig_signer_pubkeys: &[&Pubkey],
+    transfer_hook_metas: Vec<AccountMeta>,
```
buffalojoec, line 43: I know we talked about this in another comment, but I still think this should be a slice. In the case where I have no extra metas, having to allocate space for an empty vector when I can just provide an empty slice is just unnecessary, and furthermore you can just call `accounts.extend_from_slice` later in the function, which is cleaner and does the same thing. Not to mention everything else in this function signature is a reference.

program/src/offchain.rs
```diff
@@ -0,0 +1,113 @@
...
+            multisig_signer_pubkeys.is_empty(),
+        ),
+    ];
+    for signer_pubkey in multisig_signer_pubkeys.iter() {
+        accounts.push(AccountMeta::new_readonly(**signer_pubkey, true));
+    }
+
+    for meta in transfer_hook_metas.into_iter() {
+        accounts.push(meta);
+    }
```
buffalojoec, line 66: As it pertains to my other comment about the offchain module bit, this actually isn't going to add enough accounts. It needs to add all of the transfer hook metas, then the validation and hook program accounts.
https://github.com/solana-program/transfer-hook/blob/e00f3b5c591fd55b4aed6a1e9b1ccc502cb6da05/interface/src/offchain.rs#L104-L110

It's easier to just let the Token-2022 helpers do what they were designed to do.

program/tests/helpers/test-transfer-hook/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "test-transfer-hook"
+version = "0.1.0"
+edition = "2021"
+publish = false
+
+[features]
+no-entrypoint = []
+test-sbf = []
```
buffalojoec, line 9: Unused!

program/tests/helpers/test-transfer-hook/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+solana-account-info = "2.1.13"
+solana-program-entrypoint = "2.1.13"
+solana-program-error = "2.1.13"
+solana-pubkey = "2.1.13"
+
+[lib]
+crate-type = ["cdylib", "lib"]
+
+[package.metadata.docs.rs]
+targets = ["x86_64-unknown-linux-gnu"]
```
buffalojoec, line 22: Not published!

program/src/offchain.rs
grod220: Removing 👍

program/tests/helpers/test-transfer-hook/src/state.rs
```diff
@@ -0,0 +1,7 @@
+use bytemuck::{Pod, Zeroable};
+
+#[repr(transparent)]
```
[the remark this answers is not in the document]
grod220: Ah, I see what you mean. Will just use the first byte then.

scripts/rust/test-sbf.mjs
joncinque, line 1: This file should really just be for testing, and not also building the dependent programs, otherwise the point of the scripts can get confusing. We just need to be clear to run `pnpm programs:build` before `pnpm programs:test`

scripts/rust/build-sbf.mjs
joncinque, line 1: This might be ok, but note that this file doesn't really do anything anymore, and it's much easier to just run `cargo-build-sbf` instead

joncinque (commented):
I didn't look at the tests, but wanted to comment on the logic and setup of the repo

scripts/rust/build-sbf.mjs
grod220: Yes, makes sense. It was made generic so that it would build the test-hook crate as well. Think I'll just update the package.json to this and toss this file:

```
"programs:build": "cargo-build-sbf",
```

scripts/rust/test-sbf.mjs
grod220: Will revert this and add the build step to the test github action

program/src/processor.rs
```diff
@@ -309,23 +300,29 @@ pub fn process_unwrap(accounts: &[AccountInfo], amount: u64) -> ProgramResult {
...
+        recipient_unwrapped_token.clone(),
+        wrapped_mint_authority.clone(),
+    ];
+
+    for account in additional_accounts {
+        // Should filter down to only extra transfer hook accounts
+        if !multisig_signer_keys.contains(&account.key) {
+            transfer_accounts.push(account.clone());
+        }
+    }
```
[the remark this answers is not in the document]
grod220: This comment is relevant: https://github.com/solana-program/token-wrap/pull/15#discussion_r1976067475.

Passing in multisig signers to the helper makes it think it should be handled like a multisig when in this case the authority is a PDA. It's only the burn that needs multisig auth.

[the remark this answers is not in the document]
joncinque: Sorry, I don't think I understand. If you pass all the additional accounts along with the authority, then how does the helper think there's a multisig involved? `extract_multisig_accounts` will get nothing out, the authority will be identified as not-a-multisig, then the rest of the account resolution logic can play out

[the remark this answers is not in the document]
grod220: Oh wait, my brain is stuck on an earlier commit lol. You're definitely right. Working on this while mid-flight on the helper must have mixed me up. The helper PR specifically fixes this bug as it now properly identifies the multisig signers associated with the authority. Given this authority is a PDA, it will exclude them and work fine.

program/src/processor.rs
```diff
@@ -309,23 +301,23 @@ pub fn process_unwrap(accounts: &[AccountInfo], amount: u64) -> ProgramResult {
...
-            unwrapped_mint.clone(),
-            recipient_unwrapped_token.clone(),
-            wrapped_mint_authority.clone(),
-        ],
+    let mut transfer_accounts = vec![
+        unwrapped_escrow.clone(),
+        unwrapped_mint.clone(),
+        recipient_unwrapped_token.clone(),
+        wrapped_mint_authority.clone(),
+    ];
```
joncinque, line 309: Since `invoke_transfer_checked` is accepting all of these `AccountInfo`s directly, can't you just pass `additional_accounts` to `invoke_transfer_checked` as they are without creating `transfer_accounts`?

grod220: Oh! I wonder how I never noticed that 🤔

joncinque (approved):
Looks good to me! But do get an approval from Joe too

program/Cargo.toml
```diff
@@ -28,16 +28,21 @@ solana-pubkey = "2.2.1"
...
 solana-sysvar = "2.2.1"
-spl-pod = "0.5.0"
+spl-pod = "0.5.1"
 spl-token = { version = "7.0.0", features = ["no-entrypoint"] }
-spl-token-2022 = { version = "7.0.0", features = ["no-entrypoint"] }
+spl-transfer-hook-interface = "0.9.0"
 thiserror = "2.0.12"
 
+# Should depend on the next crate version after 7.0.0 when https://github.com/solana-program/token-2022/pull/233 is deployed
+spl-token-2022 = { git = "https://github.com/solana-program/token-2022", rev = "ea742a78657628e3d76755dfb38f0a4693103868", features = ["no-entrypoint"] }
```
buffalojoec, line 37: Just flagging: is this landing with this in the manifest, and being updated later, or are we halting this PR on https://github.com/solana-program/token-2022/pull/233?

program/tests/helpers/test-transfer-hook/Cargo.toml
buffalojoec, line 1: Is `tests/helpers/` the best location for a test program? Is there something more clear within `tests` we could use, like maybe `tests/programs/`?

buffalojoec (approved):
Lgtm, left a few "take it or leave it" remarks. Nice work!!

program/tests/helpers/test-transfer-hook/Cargo.toml
joncinque: I like that!

grod220: Yep, can move 👍

PR #22, opened 2025-03-04
Update deps to 2.2.1

grod220:
Required for #15


program/tests/helpers/create_mint_builder.rs
```diff
@@ -185,7 +184,7 @@ impl<'a> CreateMintBuilder<'a> {
             (wrapped_backpointer_address, wrapped_backpointer_account),
             (unwrapped_mint_addr, unwrapped_mint_account),
             (
-                system_program::id(),
+                solana_system_program::id(),
```
buffalojoec, line 187: Only if you don't need the `solana-system-program` crate for anything else:

```suggestion
                solana_sdk_ids::system_program::id(),
```

buffalojoec (commented):
Thanks for splitting this out! We were chugging along on the other PR anyway, but I appreciate the due dilligence.

joncinque (commented):
Feel free to do this in another PR, but since you're upgrading to v2.2, you might also like to upgrade `rust-toolchain.toml` to `1.84.1`, https://github.com/solana-program/token-wrap/blob/f5245a71d82aedbee3a59633aaa68a57605d6c66/Cargo.toml#L11-L12 to `nightly-2024-11-22` and https://github.com/solana-program/token-wrap/blob/f5245a71d82aedbee3a59633aaa68a57605d6c66/Cargo.toml#L6 to `2.2.0`

buffalojoec (approved):

PR #27, opened 2025-03-07
Add tests for native mint

grod220:
Ensures program successfully wraps and unwraps with native mint

### DIFF
--- a/program/tests/helpers/common.rs
+++ b/program/tests/helpers/common.rs
@@ -277,7 +277,10 @@ pub fn setup_native_token(balance: u64, owner: &TransferAuthority) -> (KeyedAcco
     state.pack_base();
 
     let native_token_account = Account {
-        lamports: Rent::default().minimum_balance(Mint::LEN),
+        lamports: Rent::default()
+            .minimum_balance(Mint::LEN)
+            .checked_add(balance)
+            .unwrap(),
         data: account_data,
         owner: spl_token_2022::id(),
         ..Default::default()
--- a/program/tests/helpers/common.rs
+++ b/program/tests/helpers/common.rs
@@ -247,3 +247,40 @@ pub fn setup_validation_state_account(
         },
     }
 }
+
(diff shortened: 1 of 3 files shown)


program/tests/helpers/common.rs
```diff
@@ -247,3 +247,40 @@ pub fn setup_validation_state_account(
...
+        amount: balance,
+        owner: owner.keyed_account.key,
+        state: AccountState::Initialized,
+        is_native: COption::Some(20),
+        ..Default::default()
+    };
+    state.pack_base();
+
+    let native_token_account = Account {
+        lamports: Rent::default().minimum_balance(Mint::LEN),
```
joncinque, line 280: nit: this should really be `Rent.minimum_balance(Account::LEN) + balance` since it's a wrapped SOL account. Its lamports should be the total amount minus the rent exempt requirement to be a token account

joncinque (approved):
Looks great, thanks for adding the test case!

program/tests/helpers/common.rs
```diff
@@ -247,3 +247,43 @@ pub fn setup_validation_state_account(
...
+        owner: owner.keyed_account.key,
+        state: AccountState::Initialized,
+        is_native: COption::Some(20),
+        ..Default::default()
+    };
+    state.pack_base();
+
+    let native_token_account = Account {
+        lamports: Rent::default()
+            .minimum_balance(Mint::LEN)
```
joncinque, line 281: This still isn't a mint :wink: 
```suggestion
            .minimum_balance(Account::LEN)
```

grod220: Whoops! Fixing.

grod220: Sometimes I get confused with the two different account types. It's `spl_token_2022::state::Account::LEN` right?

joncinque (approved):

program/tests/helpers/common.rs
joncinque: Yep you got it right!

PR #33, opened 2025-03-17
CLI helper: CreateMint

grod220:
The first of a number of CLI/JS helpers for users of the program. First up is the `CreateMint` instruction.

Takes large inspiration from the CLI helpers and tests in [spl-single-pool](https://github.com/solana-program/single-pool).

1. Build program
```bash
cargo build # builds binary
```

```bash
solana-test-validator \
  --reset
```

```bash
spl-token create-token
# Example will have G85DDQCoHchhDFWtUFT2MDkfUPYQt8BvViFyjGv167xb for this
```

```bash
    G85DDQCoHchhDFWtUFT2MDkfUPYQt8BvViFyjGv167xb \  # unwrapped_mint
    TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA \    # unwrapped_token_program
    TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb     # wrapped_token_program
```

5. See expected output
```bash
Creating wrapped mint for G85DDQCoHchhDFWtUFT2MDkfUPYQt8BvViFyjGv167xb
Funding wrapped_mint_account 7KTYY4ZAsVG2Tn1V4119ZGRLWRPzo3r6nD2pde9rtLsV with 1461600 lamports for rent
Funding backpointer_account 7ohy6W6gxAHg42pAawrUPeivRLyCBPXALgnq17eNVYFF with 1113600 lamports for rent
Unwrapped mint address: G85DDQCoHchhDFWtUFT2MDkfUPYQt8BvViFyjGv167xb
Wrapped mint address: 7KTYY4ZAsVG2Tn1V4119ZGRLWRPzo3r6nD2pde9rtLsV
Wrapped backpointer address:: 7ohy6W6gxAHg42pAawrUPeivRLyCBPXALgnq17eNVYFF
Funded wrapped mint lamports: 1461600
Funded backpointer lamports: 1113600
Signature: 3T5yjnCgDwcPAafyYrjxfFXxDZfPFRFr2R8K4kcPbZnEukKjsYxiwJrdAJAFyCU7guq5Yye5QEwobLbsMkGZNKEe
```

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -19,6 +19,12 @@ inputs:
   solana:
     description: Install Solana if `true`. Defaults to `false`.
     required: false
+  cli:
+    description: Install CLI dependencies if `true`. Defaults to `false`.
+    required: false
+  purge:
+    description: Purge unused directories if `true`. Defaults to `false`.
+    required: false
 
 runs:
   using: 'composite'
@@ -31,6 +37,25 @@ runs:
       with:
         node-version: 20
         cache: 'pnpm'
+
+    - name: Purge unused ubuntu runner directories
+      if: ${{ inputs.purge == 'true' }}
+      shell: bash
+      run: |
+        sudo rm -rf /usr/share/dotnet
+        sudo rm -rf /usr/share/swift
+        sudo rm -rf /usr/share/mysql
+        sudo rm -rf /usr/share/az_*
+        sudo rm -rf /usr/share/postgresql-common
+        sudo rm -rf /opt/ghc
+        sudo rm -rf /opt/az
+        sudo rm -rf /opt/pipx
+        sudo rm -rf /opt/microsoft
+        sudo rm -rf /opt/google
+        sudo rm -rf /opt/hostedtoolcache
+        sudo rm -rf /usr/local/lib/android
+        sudo rm -rf /usr/local/lib/heroku
+        sudo rm -rf /imagegeneration
 
     - name: Install Dependencies
       run: pnpm install --frozen-lockfile
@@ -60,6 +85,11 @@ runs:
       with:
         version: ${{ env.SOLANA_VERSION }}
         cache: true
+
+    - name: Install CLI dependencies
+      if: ${{ inputs.cli == 'true' }}
+      shell: bash
+      run: sudo apt update && sudo apt install libudev-dev protobuf-compiler -y
 
     - name: Cache Cargo Dependencies
       if: ${{ inputs.cargo-cache-key && !inputs.cargo-cache-fallback-key }}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -139,6 +139,25 @@ jobs:
       - name: Test Programs
         run: pnpm programs:test
 
+  format_and_lint_cli:
+    name: Format & Lint CLI
+    runs-on: ubuntu-latest
+    steps:
+      - name: Git Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Environment
+        uses: ./.github/actions/setup
+        with:
+          clippy: true
+          rustfmt: true
+
+      - name: Format CLI
+        run: pnpm clients:cli:format
+
+      - name: Lint CLI
+        run: pnpm clients:cli:lint
+
   test_client_cli:
     name: Test Client CLI
     runs-on: ubuntu-latest
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -149,6 +149,9 @@ jobs:
       - name: Setup Environment
         uses: ./.github/actions/setup
         with:
+          cargo-cache-key: cargo-cli
+          cli: true
+          purge: true
           clippy: true
           rustfmt: true
 
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,9 +2,9 @@ name: Main
 
 on:
   push:
-    branches: [main]
+    branches: [ main ]
   pull_request:
-    branches: [main]
+    branches: [ main ]
 
 jobs:
   format_and_lint_programs:
@@ -138,3 +138,34 @@ jobs:
 
       - name: Test Programs
         run: pnpm programs:test
+
+  test_client_cli:
+    name: Test Client CLI
+    runs-on: ubuntu-latest
+    needs: build_programs
+    steps:
+      - name: Git Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Environment
+        uses: ./.github/actions/setup
+        with:
+          cargo-cache-key: cargo-cli
+          solana: true
+          cli: true
+          purge: true
+
+      - name: Restore Program Builds
+        uses: actions/cache/restore@v4
+        with:
+          path: ./**/*.so
+          key: ${{ runner.os }}-builds-${{ github.sha }}
+
+      - name: Build Programs
+        run: pnpm programs:build
+
+      - name: Build Binary
+        run: cargo build
+
+      - name: Test
+        run: pnpm clients:cli:test
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -151,7 +151,6 @@ jobs:
         uses: ./.github/actions/setup
         with:
           cargo-cache-key: cargo-cli
-          solana: true
           cli: true
           purge: true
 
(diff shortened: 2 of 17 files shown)


.github/actions/setup/action.yml
```diff
@@ -32,6 +38,25 @@ runs:
...
+        sudo rm -rf /usr/share/postgresql-common
+        sudo rm -rf /opt/ghc
+        sudo rm -rf /opt/az
+        sudo rm -rf /opt/pipx
+        sudo rm -rf /opt/microsoft
+        sudo rm -rf /opt/google
+        sudo rm -rf /opt/hostedtoolcache
+        sudo rm -rf /usr/local/lib/android
+        sudo rm -rf /usr/local/lib/heroku
+        sudo rm -rf /imagegeneration
```
grod220, line 58: Taken from https://github.com/solana-program/single-pool/blob/main/.github/actions/setup/action.yml#L41-L58. Assuming it's required in order to get solana test validator working.

clients/cli/src/create_mint.rs
```diff
@@ -0,0 +1,215 @@
...
+
+    let mint_rent = rpc_client
+        .get_minimum_balance_for_rent_exemption(spl_token_2022::state::Mint::LEN)
+        .await?;
+
+    let mut funded_wrapped_mint_lamports = 0;
+    if wrapped_mint_lamports < mint_rent {
+        let lamports_to_transfer = mint_rent
+            .checked_sub(wrapped_mint_lamports)
+            .ok_or(anyhow!("Error subtracting wrapped_mint_lamports from rent"))?;
```
buffalojoec, line 135: You already checked to make sure `wrapped_mint_lamports` is less than `mint_rent`, so you could use `saturating_sub` here if you want.

clients/cli/src/create_mint.rs
```diff
@@ -0,0 +1,215 @@
...
+        Err(_) => 0,
+    };
+
+    let mint_rent = rpc_client
+        .get_minimum_balance_for_rent_exemption(spl_token_2022::state::Mint::LEN)
+        .await?;
+
+    let mut funded_wrapped_mint_lamports = 0;
+    if wrapped_mint_lamports < mint_rent {
+        let lamports_to_transfer = mint_rent
```
buffalojoec, line 133: Up to you, but you could just say

```rust
let funded_wrapped_mint_lamports = mint_rent.saturating_sub(wrapped_mint_lamports);
```

and then just add the instruction if `funded_wrapped_mint_lamports > 0`.

clients/cli/src/create_mint.rs
```diff
@@ -0,0 +1,215 @@
...
+    };
+
+    let backpointer_rent = rpc_client
+        .get_minimum_balance_for_rent_exemption(std::mem::size_of::<
+            spl_token_wrap::state::Backpointer,
+        >())
+        .await?;
+
+    let mut funded_backpointer_lamports = 0;
+    if backpointer_lamports < backpointer_rent {
```
buffalojoec, line 165: Same cleanup suggestion applies here, if you like it!

clients/cli/src/config.rs
```diff
@@ -0,0 +1,89 @@
...
+    solana_cli_output::OutputFormat,
+    solana_client::nonblocking::rpc_client::RpcClient,
+    solana_commitment_config::CommitmentConfig,
+    solana_remote_wallet::remote_wallet::RemoteWalletManager,
+    solana_signer::Signer,
+    std::{rc::Rc, sync::Arc},
+};
+
+pub struct Config {
+    pub rpc_client: Arc<RpcClient>,
```
buffalojoec, line 15: @joncinque I don't think I understand the relationship between `rpc_client` and `program_client` in other CLIs like the Single Pool one. But, if we don't need it here - which it looks like we don't - then I'm good with this!

buffalojoec (commented):
Really small stuff from me. Looks great!

joncinque (commented):
Looks great overall! Mostly small things

scripts/rust/test.mjs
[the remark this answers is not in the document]
grod220: We should brainstorm some better patterns!

clients/cli/tests/helpers.rs
```diff
@@ -0,0 +1,123 @@
...
+    spl_token::{self, instruction::initialize_mint, state::Mint as SplTokenMin},
+    spl_token_client::client::{ProgramClient, ProgramRpcClient, ProgramRpcClientSendTransaction},
+    spl_token_wrap::{self},
+    std::{path::PathBuf, sync::Arc},
+    tempfile::NamedTempFile,
+};
+
+pub const TOKEN_WRAP_CLI_BIN: &str = "../../target/debug/spl-token-wrap";
+
+pub type PClient = Arc<dyn ProgramClient<ProgramRpcClientSendTransaction>>;
```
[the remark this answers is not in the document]
grod220: Ah true, removed.

clients/cli/Cargo.toml
```diff
@@ -0,0 +1,48 @@
+[package]
+name = "spl-token-wrap-cli"
+version = "1.0.0"
```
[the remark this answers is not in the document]
buffalojoec: It doesn't matter all that much since we wouldn't publish until it's all finished, but I was just curious if you wanted to start with `0.1.0` instead, since it gives us a few minor releases before the first major. Depends what the roadmap looks like for this thing.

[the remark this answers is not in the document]
buffalojoec: I like to start with `0.0.1` so I can break it a bunch. 😅

clients/cli/tests/test_create_mint.rs
```diff
@@ -0,0 +1,67 @@
...
+mod helpers;
+
+#[tokio::test]
+#[serial]
+async fn test_create_mint() {
+    let env = setup().await;
+    let unwrapped_token_program = spl_token::id();
+    let wrapped_token_program = spl_token_2022::id();
+    let unwrapped_mint =
+        create_unwrapped_mint(&env.rpc_client, &env.payer, &unwrapped_token_program).await;
```
buffalojoec, line 22: This is totally up to you, and sorry I didn't spot this earlier, but you could let `setup()` and/or `start_validator` take some flavor of a `(Pubkey, Account)` input parameter, and boot the test validator with those accounts already present. This way, you wouldn't need to invoke the RPC to create an unwrapped mint. You just start the validator and the mint exists already.

buffalojoec (commented):
Lgtm, just a CI failure to address!

grod220:
Since swapping from ProgramClient to RpcClient in the tests I am hitting flakiness with a ~30% success rate on the cli test. I think it may be because the ProgramClient perhaps waits for confirmation and the other doesn't. Not sure, investigating.

buffalojoec (approved):

.github/workflows/main.yml
```diff
@@ -138,3 +138,27 @@ jobs:
 
       - name: Test Programs
         run: pnpm programs:test
+
+  test_client_cli:
```
joncinque, line 142: Can you also add lint + format checks on the cli package?

joncinque (commented):
When using `rpc_client.send_transaction`, it *only* sends.

So rather than setting all of the data at startup to "fix" the race condition, the preference should be to use `rpc_client.send_and_confirm_transaction` to wait for a transaction to land before sending the next one, if the two transactions depend on each other.

You don't need to change setting the accounts into the validator, but can you at least use `send_and_confirm_transaction` in the tests? That also lets us find out if the transaction was successful or not.

Also, it's done now, but the workspace deps changes should have been done in a separate PR.

grod220:
> the preference should be to use rpc_client.send_and_confirm_transaction to wait for a transaction to land before sending the next one, if the two transactions depend on each other.

When digging into the ProgramClient, I did manage to find `send_and_confirm_transaction` which fixed the issue. However, if all accounts are loaded on startup, that method isn't needed anymore and there is only the single command now:
```
 Command::new(TOKEN_WRAP_CLI_BIN).args(["create-mint", ...
```
Though it's still not clear to me what strategy is preferred for CLI tests. Is it better to A) load all accounts within the validator on startup and make the single command you are testing or B) have a fresh startup state and make the sequential transactions to build the state the command your testing needs. If A, there shouldn't be additional changes needed. If B, I can revert back and make use of `send_and_confirm_transaction` to fix that race condition.

joncinque (approved):
I typically lean towards building things up in tests with transactions rather than setting up the accounts by hand, since it's possible for the setup to be incorrect, as we've already seen in some of the program tests earlier, but there's no need to revert the work here.

And right, sorry, I thought there was still some usage of `send_transaction` somewhere. This is all good to me then!

grod220:
Helpful, in the next PR will probably re-establish the multiple-tx pattern again.

PR #34, opened 2025-03-18
Client helpers: CreateMint (javascript)

grod220:
Creating a js client library for calling the `CreateMint` instruction (1 of 3 total). Attempts to use solana kit as the base library. 

1. Build program
```bash
cargo build # builds binary
```

```bash
solana-test-validator \
  --reset
```

```bash
spl-token create-token
# Example will have G85DDQCoHchhDFWtUFT2MDkfUPYQt8BvViFyjGv167xb for this
```

4. Update `token-wrap/clients/js/example.ts` with your keypair + unwrapped mint address generated above

5. Install deps
```bash
cd ./token-wrap/clients/js
pnpm install
```

6. Run example
```bash
pnpm example:create-mint
```

7. See expected output
```bash
Wrapped Mint: 4yvMNhbXqbZC7oYrgNY3dYhsRsGUK4Cdsz5RVwDBJTnH
Backpointer: A4hV7H5NY6xNSfeWa1goAJeEkRYrctUFoLjiThDfTJN7
Funded wrapped mint lamports: 1461600n
Funded backpointer lamports: 1113600n
Signature: 47DidWXkH62Bzj8bq3fpN737ofUXUvZGq2U2U9oTQuQKgmDS2mnyHGmatEWN2Rjwj8ADwqNShsS8nbQQeAphVyU4
```


clients/js/src/create-mint.ts
```diff
@@ -0,0 +1,131 @@
...
+        amount: fundedWrappedMintLamports,
+      }),
+    );
+  }
+
+  // Fund backpointer account if needed
+  let fundedBackpointerLamports = 0n;
+  const backpointerAccount = await rpc.getAccountInfo(backpointer).send();
+  const backpointerSize = BigInt(getBackpointerSize());
+  const backpointerRent = await rpc.getMinimumBalanceForRentExemption(backpointerSize).send();
```
lorisleiva, line 72: Here, I'd put the two await in a `Promise.all` since they can be executed in parallel.

clients/js/src/create-mint.ts
```diff
@@ -0,0 +1,131 @@
...
+    {
+      unwrappedMint,
+      wrappedTokenProgram: wrappedTokenProgram,
+    },
+    { programAddress: TOKEN_WRAP_PROGRAM_ADDRESS },
+  );
+  const [backpointer] = await findBackpointerPda(
+    { wrappedMint },
+    { programAddress: TOKEN_WRAP_PROGRAM_ADDRESS },
+  );
```
lorisleiva, line 47: I believe the `programAddress` option is redundant since the PDA is generated from that program anyway.

```suggestion
  const [wrappedMint] = await findWrappedMintPda({
    unwrappedMint,
    wrappedTokenProgram: wrappedTokenProgram,
  });
  const [backpointer] = await findBackpointerPda({ wrappedMint });
```

clients/js/src/create-mint.ts
```diff
@@ -0,0 +1,131 @@
...
+      idempotent,
+    }),
+  );
+
+  const { value: latestBlockhash } = await rpc.getLatestBlockhash().send();
+
+  // Build transaction
+  let tx = pipe(
+    createTransactionMessage({ version: 0 }),
+    tx => setTransactionMessageFeePayer(payer.address, tx),
```
lorisleiva, line 101: You have a Signer, you can leverage it.

```suggestion
    tx => setTransactionMessageFeePayerSigner(payer, tx),
```

clients/js/src/create-mint.ts
```diff
@@ -0,0 +1,131 @@
...
+  if (!signatures) {
+    throw new Error('Expected a signature for compiled transaction');
+  }
+
+  const wireFormatTransaction = getBase64EncodedWireTransaction({
+    ...compiledTransaction,
+    signatures,
+  });
+
+  const signature = await rpc.sendTransaction(wireFormatTransaction, { encoding: 'base64' }).send();
```
lorisleiva, line 122: All this can be replaced with:

```suggestion
   // You passed signers everywhere so we know who needs to sign the transaction and how.
  const signedTransaction = await signTransactionMessageWithSigners(tx);
  
  // Create a send and confirm function from your RPC and RPC Subscriptions objects.
  // You can create the latter using `createSolanaRpcSubscriptions`.
  const sendAndConfirm = sendAndConfirmTransactionFactory({ rpc, rpcSubscriptions });
 
  // Use it to send and confirm your signed transaction.
  await sendAndConfirm(signedTransaction, { commitment: 'confirmed' });
  
  // Access the signature of your signed transaction (this can be done even before sending it).
  const signature = getSignatureFromTransaction(signedTransaction);
```

clients/js/src/example.ts
```diff
@@ -0,0 +1,37 @@
...
+  58, 188, 194, 176, 230, 94, 253, 2, 24, 163, 198, 177, 92, 79, 213, 87, 122, 150, 216, 175, 176,
+  159, 113, 144, 148, 82, 149, 249, 242, 255, 7, 1, 73, 203, 66, 98, 4, 2, 141, 236, 49, 10, 47,
+  188, 93, 170, 111, 125, 44, 155, 4, 124, 48, 18, 188, 30, 158, 78, 158, 34, 44, 100, 61, 21,
+]);
+const UNWRAPPED_MINT_ADDRESS = address('5HXwCPsqa8cZSAXDimAW9vJB8b3VdjCMWt1aLrCT2Wpb');
+
+const main = async () => {
+  const rpc = createSolanaRpc('http://127.0.0.1:8899/');
+  const keyPair = await createKeyPairFromBytes(PRIVATE_KEY_PAIR);
+  const payer = await createSignerFromKeyPair(keyPair);
```
lorisleiva, line 21: nit: Promise.all

lorisleiva (commented):
I'm impressed! I've made some comments/suggestions on the JS client. Let me know what you think.

clients/js/src/example.ts
grod220: Since keyPair is a parameter of the next async call, they have to be sequential

clients/js/src/create-mint.ts
```diff
@@ -0,0 +1,131 @@
...
+        source: payer,
+        destination: wrappedMint,
+        amount: fundedWrappedMintLamports,
+      }),
+    );
+  }
+
+  // Fund backpointer account if needed
+  let fundedBackpointerLamports = 0n;
+  const backpointerAccount = await rpc.getAccountInfo(backpointer).send();
```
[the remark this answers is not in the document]
grod220: Given that these accounts need to be prefunded (if they don't already exist), I can't assert, but can do something like:

```typescript
const wrappedMintLamports = wrappedMintAccount.exists ? wrappedMintAccount.lamports : 0n;
```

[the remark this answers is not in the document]
lorisleiva: Perfect 👌

clients/js/src/example.ts
lorisleiva: Ah yeah sorry I missed that. 🙂

clients/js/package.json
```diff
@@ -0,0 +1,63 @@
...
+  },
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/solana-program/token-wrap.git"
+  },
+  "bugs": {
+    "url": "https://github.com/solana-program/token-wrap/issues"
+  },
+  "peerDependencies": {
+    "@solana/kit": "^2.1.0"
```
lorisleiva, line 44: You should also add this to `devDependencies` otherwise you won't have it installed locally. Whilst there is no JS tests for this client at the moment, running the example script should also fail. My guess is this is working for you locally right now because the `node_modules` folder still has is from before this change.

clients/js/package.json
```diff
@@ -0,0 +1,63 @@
...
+  },
+  "dependencies": {
+    "@solana-program/token-2022": "^0.4.0"
```
lorisleiva, line 47: I see you also use `@solana-program/system` in the `create-mint.ts` file.

clients/js/src/example.ts
```diff
@@ -0,0 +1,40 @@
...
+  159, 113, 144, 148, 82, 149, 249, 242, 255, 7, 1, 73, 203, 66, 98, 4, 2, 141, 236, 49, 10, 47,
+  188, 93, 170, 111, 125, 44, 155, 4, 124, 48, 18, 188, 30, 158, 78, 158, 34, 44, 100, 61, 21,
+]);
+const UNWRAPPED_MINT_ADDRESS = address('5HXwCPsqa8cZSAXDimAW9vJB8b3VdjCMWt1aLrCT2Wpb');
+
+const main = async () => {
+  const rpc = createSolanaRpc('http://127.0.0.1:8899');
+  const rpcSubscriptions = createSolanaRpcSubscriptions('ws://127.0.0.1:8900');
+  const keyPair = await createKeyPairFromBytes(PRIVATE_KEY_PAIR);
+  const payer = await createSignerFromKeyPair(keyPair);
```
lorisleiva, line 23: nit: there's actually a helper that does both. Sorry I missed that before.

```suggestion
  const payer = await createKeyPairSignerFromBytes(PRIVATE_KEY_PAIR);
```

clients/js/src/global.d.ts
```diff
@@ -0,0 +1,8 @@
+// Done to appease issue in generated code:
+//  error TS4111: Property 'NODE_ENV' comes from an index signature, so it must be accessed with ['NODE_ENV'].
+
+declare namespace NodeJS {
+  interface ProcessEnv {
+    NODE_ENV: string;
+  }
+}
```
lorisleiva, line 8: Interesting, I wonder if it'd be safer for Codama to generate as `['NODE_ENV']` then.

I added an issue on Codama (https://github.com/codama-idl/codama/issues/523) to discuss this.

lorisleiva (approved):
Very nice! Just a few last comments/suggestions but it's good to merge after that. 👌

PR #38, opened 2025-03-24
CLI helper: Wrap

grod220:
The second CLI helper (after https://github.com/solana-program/token-wrap/pull/33) for the `Wrap` instruction. Note, it only supports single signer at the moment. 

1. Build program
```bash
cargo build # builds binary
```

```bash
solana-test-validator \
  --reset
```

```bash
spl-token create-token
# mint addr -> G85DDQCoHchhDFWtUFT2MDkfUPYQt8BvViFyjGv167xb
# token account -> EYUCh7xP4sxnNBr1VFcGXfHm1piEA6ERci5prNykSV4G
```

```bash
    G85DDQCoHchhDFWtUFT2MDkfUPYQt8BvViFyjGv167xb \  # unwrapped_mint
    TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA \    # unwrapped_token_program
    TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb     # wrapped_token_program

# Keep note of console output
# Unwrapped mint address: 8eZ2rk3GcKopnKrAmVibV7zLW1sszNo2GdbhtxRhrZRi
# Wrapped mint address: 3xnGREX1Mi7zcuXJAYoDGk9xjpugZnfRhAKLPHdfEPPq
# Wrapped mint authority: A2LJVqpdm8kQUpw9x6PBZNVKzG9KesxF7gpbbvHxe8m
```

5. Create an escrow token account and set its owner to the mint_authority
```bash
spl-token create-account 8eZ2rk3GcKopnKrAmVibV7zLW1sszNo2GdbhtxRhrZRi --owner A2LJVqpdm8kQUpw9x6PBZNVKzG9KesxF7gpbbvHxe8m --fee-payer /Users/gabe/.config/solana/id.json
```

6. Create a recipient token account of that new wrapped mint
```bash
spl-token create-account 3xnGREX1Mi7zcuXJAYoDGk9xjpugZnfRhAKLPHdfEPPq
```

7. Issue Wrap command
```bash
spl-token-wrap wrap <UNWRAPPED_TOKEN_PROGRAM> <UNWRAPPED_MINT> <UNWRAPPED_TOKEN_ACCOUNT> <ESCROW_ACCOUNT> <WRAPPED_TOKEN_PROGRAM> <RECIPIENT_TOKEN_ACCOUNT> <AMOUNT>

cargo run --bin spl-token-wrap wrap TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA 8eZ2rk3GcKopnKrAmVibV7zLW1sszNo2GdbhtxRhrZRi EYUCh7xP4sxnNBr1VFcGXfHm1piEA6ERci5prNykSV4G 27QMHK3W6qYni1abKz7tK7A6Sm69jvLaBv8zataRppEy TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb 6v5uYPBuubQqDwFU2uT6uQYC4P89H8iHzJgA4q7pvPHR 55

```

9. See console output
```bash
Wrapping 55 tokens from mint 8eZ2rk3GcKopnKrAmVibV7zLW1sszNo2GdbhtxRhrZRi
Unwrapped mint address: 8eZ2rk3GcKopnKrAmVibV7zLW1sszNo2GdbhtxRhrZRi
Wrapped mint address: 3xnGREX1Mi7zcuXJAYoDGk9xjpugZnfRhAKLPHdfEPPq
Unwrapped token account: EYUCh7xP4sxnNBr1VFcGXfHm1piEA6ERci5prNykSV4G
Recipient wrapped token account: 6v5uYPBuubQqDwFU2uT6uQYC4P89H8iHzJgA4q7pvPHR
Escrow account: 27QMHK3W6qYni1abKz7tK7A6Sm69jvLaBv8zataRppEy
Amount: 55
Signature: FRbCJmutYYFiGtnXzMFSgHucMRpdkHBHh4SCPX984qVQZRTfsFkvyKnnkbjW7M1nw1ZAKiTBC1igMnRSGg1H4x4
```

### DIFF
--- a/clients/cli/Cargo.toml
+++ b/clients/cli/Cargo.toml
@@ -13,7 +13,6 @@ clap = { workspace = true }
 serde = { workspace = true }
 serde_derive = { workspace = true }
 serde_with = { workspace = true }
-solana-account = { workspace = true }
 solana-clap-v3-utils = { workspace = true }
 solana-cli-config = { workspace = true }
 solana-cli-output = { workspace = true }
--- a/clients/cli/Cargo.toml
+++ b/clients/cli/Cargo.toml
@@ -27,7 +27,7 @@ solana-signature = { workspace = true }
 solana-signer = { workspace = true }
 solana-system-interface = { workspace = true }
 solana-transaction = { workspace = true }
-spl-associated-token-account = { workspace = true }
+spl-associated-token-account-client = { workspace = true }
 spl-token = { workspace = true }
 spl-token-wrap = { workspace = true }
 spl-token-2022 = { workspace = true }
--- a/clients/cli/Cargo.toml
+++ b/clients/cli/Cargo.toml
@@ -27,6 +27,7 @@ solana-signature = { workspace = true }
 solana-signer = { workspace = true }
 solana-system-interface = { workspace = true }
 solana-transaction = { workspace = true }
+spl-associated-token-account = { workspace = true }
 spl-token = { workspace = true }
 spl-token-wrap = { workspace = true }
 spl-token-2022 = { workspace = true }
--- a/clients/cli/src/create_mint.rs
+++ b/clients/cli/src/create_mint.rs
@@ -16,8 +16,8 @@ use {
     solana_system_interface::instruction::transfer,
     solana_transaction::Transaction,
     spl_token_wrap::{
-        get_wrapped_mint_address, get_wrapped_mint_backpointer_address, id,
-        instruction::create_mint,
+        get_wrapped_mint_address, get_wrapped_mint_authority, get_wrapped_mint_backpointer_address,
+        id, instruction::create_mint,
     },
     std::fmt::{Display, Formatter},
 };
@@ -50,6 +50,8 @@ pub struct CreateMintOutput {
     #[serde_as(as = "DisplayFromStr")]
     pub wrapped_mint_address: Pubkey,
     #[serde_as(as = "DisplayFromStr")]
+    pub wrapped_mint_authority: Pubkey,
+    #[serde_as(as = "DisplayFromStr")]
     pub wrapped_backpointer_address: Pubkey,
     pub funded_wrapped_mint_lamports: u64,
     pub funded_backpointer_lamports: u64,
@@ -68,6 +70,11 @@ impl Display for CreateMintOutput {
             f,
             "Wrapped mint address:",
             &self.wrapped_mint_address.to_string(),
+        )?;
+        writeln_name_value(
+            f,
+            "Wrapped mint authority:",
+            &self.wrapped_mint_authority.to_string(),
         )?;
         writeln_name_value(
             f,
@@ -196,6 +203,7 @@ pub async fn command_create_mint(config: &Config, args: CreateMintArgs) -> Comma
         CreateMintOutput {
             unwrapped_mint_address: args.unwrapped_mint,
             wrapped_mint_address,
+            wrapped_mint_authority: get_wrapped_mint_authority(&wrapped_mint_address),
             wrapped_backpointer_address,
             funded_wrapped_mint_lamports,
             funded_backpointer_lamports,
--- a/clients/cli/src/create_mint.rs
+++ b/clients/cli/src/create_mint.rs
@@ -16,8 +16,8 @@ use {
     solana_system_interface::instruction::transfer,
     solana_transaction::Transaction,
     spl_token_wrap::{
-        get_wrapped_mint_address, get_wrapped_mint_authority, get_wrapped_mint_backpointer_address,
-        id, instruction::create_mint,
+        get_wrapped_mint_address, get_wrapped_mint_backpointer_address, id,
+        instruction::create_mint,
     },
     std::fmt::{Display, Formatter},
 };
@@ -50,8 +50,6 @@ pub struct CreateMintOutput {
     #[serde_as(as = "DisplayFromStr")]
     pub wrapped_mint_address: Pubkey,
     #[serde_as(as = "DisplayFromStr")]
-    pub wrapped_mint_authority: Pubkey,
-    #[serde_as(as = "DisplayFromStr")]
     pub wrapped_backpointer_address: Pubkey,
     pub funded_wrapped_mint_lamports: u64,
     pub funded_backpointer_lamports: u64,
@@ -70,11 +68,6 @@ impl Display for CreateMintOutput {
             f,
             "Wrapped mint address:",
             &self.wrapped_mint_address.to_string(),
-        )?;
-        writeln_name_value(
-            f,
-            "Wrapped mint authority:",
-            &self.wrapped_mint_authority.to_string(),
         )?;
         writeln_name_value(
             f,
@@ -203,7 +196,6 @@ pub async fn command_create_mint(config: &Config, args: CreateMintArgs) -> Comma
         CreateMintOutput {
             unwrapped_mint_address: args.unwrapped_mint,
             wrapped_mint_address,
-            wrapped_mint_authority: get_wrapped_mint_authority(&wrapped_mint_address),
             wrapped_backpointer_address,
             funded_wrapped_mint_lamports,
             funded_backpointer_lamports,
--- a/clients/cli/src/wrap.rs
+++ b/clients/cli/src/wrap.rs
@@ -3,17 +3,20 @@ use {
         common::{parse_pubkey, parse_token_program, process_transaction},
         config::Config,
         output::{format_output, println_display},
-        CommandResult,
+        CommandResult, Error,
     },
     clap::Args,
     serde_derive::{Deserialize, Serialize},
     serde_with::{serde_as, DisplayFromStr},
     solana_cli_output::{display::writeln_name_value, QuietDisplay, VerboseDisplay},
+    solana_client::nonblocking::rpc_client::RpcClient,
     solana_pubkey::Pubkey,
     solana_remote_wallet::remote_wallet::RemoteWalletManager,
     solana_signature::Signature,
     solana_signer::Signer,
     solana_transaction::Transaction,
+    spl_associated_token_account::get_associated_token_address_with_program_id,
+    spl_token_2022::{extension::PodStateWithExtensions, pod::PodAccount},
     spl_token_wrap::{get_wrapped_mint_address, get_wrapped_mint_authority, instruction::wrap},
     std::{
         fmt::{Display, Formatter},
@@ -28,10 +31,6 @@ pub struct WrapArgs {
     #[clap(value_parser = parse_token_program)]
     pub unwrapped_token_program: Pubkey,
 
-    /// The address of the mint to wrap
-    #[clap(value_parser = parse_pubkey)]
-    pub unwrapped_mint: Pubkey,
-
     /// The address of the unwrapped token account to wrap from
     #[clap(value_parser = parse_pubkey)]
     pub unwrapped_token_account: Pubkey,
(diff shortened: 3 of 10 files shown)


clients/cli/tests/helpers.rs
```diff
@@ -75,36 +62,237 @@ pub async fn setup() -> Env {
...
-    let lamports = Rent::default().minimum_balance(data.len());
+    let blockhash = rpc_client.get_latest_blockhash().await.unwrap();
 
-    let account = Account {
-        lamports,
-        data,
-        owner: UNWRAPPED_TOKEN_PROGRAM,
-        ..Default::default()
-    };
-    (Pubkey::new_unique(), account.into())
```
grod220, line 109: Taking action on some feedback in last PR: https://github.com/solana-program/token-wrap/pull/33#pullrequestreview-2707626189

clients/cli/src/create_mint.rs
```diff
@@ -50,6 +50,8 @@ pub struct CreateMintOutput {
     #[serde_as(as = "DisplayFromStr")]
     pub wrapped_mint_address: Pubkey,
     #[serde_as(as = "DisplayFromStr")]
+    pub wrapped_mint_authority: Pubkey,
```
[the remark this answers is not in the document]
buffalojoec: Do you need this in production, or only tests? Because it looks to me like you're deriving it.

clients/cli/src/wrap.rs
```diff
@@ -0,0 +1,196 @@
...
+    /// The address of the token program that the wrapped mint should belong to
+    #[clap(value_parser = parse_token_program)]
+    pub wrapped_token_program: Pubkey,
+
+    /// The address of the token account to receive wrapped tokens
+    #[clap(value_parser = parse_pubkey)]
+    pub recipient_token_account: Pubkey,
+
+    /// The amount of tokens to wrap
+    #[clap(value_parser = |s: &str| -> Result<u64, String> { s.parse::<u64>().map_err(|e| e.to_string()) })]
```
buffalojoec, line 52: You shouldn't need to be this explicit, right?

```suggestion
    #[clap(value_parser)]
```

clients/cli/tests/helpers.rs
```diff
@@ -75,36 +62,237 @@ pub async fn setup() -> Env {
...
-    let state = spl_token::state::Mint {
-        decimals: 8,
-        is_initialized: true,
-        supply: 1_000_000_000,
-        mint_authority: COption::Some(*mint_authority),
-        freeze_authority: COption::None,
-    };
-    let mut data = vec![0u8; spl_token::state::Mint::LEN];
-    state.pack_into_slice(&mut data);
+pub async fn create_unwrapped_mint(
```
buffalojoec, line 75: For what it's worth, I prefer the startup account fixtures approach you had here originally. I saw the back-and-forth [over here](https://github.com/solana-program/token-wrap/pull/33#pullrequestreview-2707626189), so I'll leave it up to you to choose which you prefer, and it seems you have.

buffalojoec (commented):
Looking great, just a few questions from my side, really.

clients/cli/src/wrap.rs
```diff
@@ -0,0 +1,196 @@
...
+    #[clap(value_parser = parse_pubkey)]
+    pub escrow_account: Pubkey,
+
+    /// The address of the token program that the wrapped mint should belong to
+    #[clap(value_parser = parse_token_program)]
+    pub wrapped_token_program: Pubkey,
+
+    /// The address of the token account to receive wrapped tokens
+    #[clap(value_parser = parse_pubkey)]
+    pub recipient_token_account: Pubkey,
```
joncinque, line 49: This could also be optional, and default to the ATA for the fee payer

joncinque (commented):
Looks great overall! Mostly small things

clients/cli/tests/helpers.rs
buffalojoec: > Do you have a reason for that preference?

I have formed a bias toward fixtures for a number of reasons, but it spawned from unit testing the RPC API for `@solana/kit` and has been hardened through writing program unit tests with Mollusk. It was also the motivation for building the latter.

The crux of the fixtures/mocks approach is the belief that your unit tests should be testing a single unit in isolation. If you're trying to test function C and its setup depends on functions A and B, now your tests for C are all coupled to functions A and B.

What if, all this time, `CreateMint` was doing something slightly erroneous, and we were getting _all_ false positives in `Wrap` and `Unwrap`? It might sound ridiculous, but it happens.

Fixtures/mocks also let you set up a wide array of edge cases, which might be complex to set up through transactions (perhaps they would require a custom program, for example).

I also think if we're going to write unit tests that use separate test-validator instances, the faster they spin up and then back down, the better. That means less "send and confirm".

> We can certainly stick with fixtures for unit tests, but for a CLI, an end-to-end test is more useful to test the whole flow, but maybe I'm missing some context.

I suppose I would just set up the per-command unit tests in isolation and then have one or more aptly-labeled end-to-end tests for the whole flow. I wouldn't make every CLI test an end-to-end test.

> We've had false positives in the past due to improper setup in tests, so I lean towards the more complicated setup to be sure we're actually testing a real-world situation.

Without knowing the context of these false positives I don't want to make any assumptions, but I generally think this is an acceptable risk compared to coupling too many things within each test.

Anyway, that's just my two cents! If you both disagree with me, you can do what you think is best!

joncinque: In my experience, there's a much higher chance of false positives with unit-testing than end-to-end testing because you can setup any set of inputs. As one example, remember all of the firedancer conformance fixtures that were checking impossible situations?

On the flip side, I'll definitely admit that highly-coupled tests are a lot more work to maintain. But I still stick by these words: [Write tests. Not too many. Mostly integration](https://kentcdodds.com/blog/write-tests)

For now, since we don't have integration tests otherwise, I lean towards doing them here in the CLI. But we can let Gabe be the tie-breaker :sweat_smile:

clients/cli/src/wrap.rs
grod220: Added this functionality and a test for it.

clients/cli/tests/helpers.rs
grod220: Definitely can see the two sides to this. One benefit I've seen with these CLI tests making the actual calls is that it's shown me the real-life process the end-user will need to go through in order to make use of it. This inspired https://github.com/solana-program/token-wrap/pull/40 because I ran into a user flow I wasn't fully supporting.

One thing to note is in the main program code with mollusk, the processor methods were tested in isolation with pre-made accounts passed in, which feels like it worked well given the variety of cases those tests covered. Given that, I think this program probably could use the "integration coverage" by continuing along the path of this PR. Will go that route for these CLI tests.

clients/cli/Cargo.toml
```diff
@@ -28,6 +27,7 @@ solana-signature = { workspace = true }
 solana-signer = { workspace = true }
 solana-system-interface = { workspace = true }
 solana-transaction = { workspace = true }
+spl-associated-token-account = { workspace = true }
```
joncinque, line 30: nit: can be down in follow-up work but `spl-associated-token-account-client` https://crates.io/crates/spl-associated-token-account-client/2.0.0 will give you what you need with fewer deps

(also, doesn't matter since the dep will change, but if using `spl-associated-token-account`, the `no-entrypoint` feature should also be enabled)

clients/cli/src/wrap.rs
```diff
@@ -0,0 +1,224 @@
...
+    #[clap(value_parser)]
+    pub amount: u64,
+
+    /// Path to the signer for the transfer authority if different from
+    /// fee payer
+    #[clap(long = "transfer-authority", value_name = "PATH")]
+    pub transfer_authority: Option<String>,
+
+    /// The address of the mint to wrap, queried if not provided
+    #[clap(long = "unwrapped-mint", value_parser = parse_pubkey)]
```
joncinque, line 56: nit: can remove the name here, since clap figures it out automatically
```suggestion
    #[clap(long, value_parser = parse_pubkey)]
```

clients/cli/src/wrap.rs
```diff
@@ -0,0 +1,224 @@
...
+    pub unwrapped_mint: Option<Pubkey>,
+
+    /// The address of the token account to receive wrapped tokens.
+    /// If not provided, defaults to fee payer associated token account
+    #[clap(long = "recipient-token-account", value_parser = parse_pubkey)]
```
joncinque, line 61: nit: same here, let clap figure it out automatically
```suggestion
    #[clap(long, value_parser = parse_pubkey)]
```

clients/cli/src/wrap.rs
```diff
@@ -0,0 +1,224 @@
...
+    #[clap(value_parser = parse_token_program)]
+    pub wrapped_token_program: Pubkey,
+
+    /// The amount of tokens to wrap
+    #[clap(value_parser)]
+    pub amount: u64,
+
+    /// Path to the signer for the transfer authority if different from
+    /// fee payer
+    #[clap(long = "transfer-authority", value_name = "PATH")]
```
joncinque, line 52: nit: same here, let clap figure it out automatically
```suggestion
    #[clap(long, value_name = "PATH")]
```

clients/cli/src/wrap.rs
```diff
@@ -0,0 +1,224 @@
...
+
+impl QuietDisplay for WrapOutput {
+    fn write_str(&self, _: &mut dyn std::fmt::Write) -> std::fmt::Result {
+        Ok(())
+    }
+}
+impl VerboseDisplay for WrapOutput {}
+
+async fn get_unwrapped_mint(
+    rpc_client: &Arc<RpcClient>,
```
joncinque, line 123: nit: passing in a `&Arc<T>` is an antipattern if the Arc isn't needed, it's simpler and clearer to just take in the ref directly
```suggestion
    rpc_client: &RpcClient,
```

clients/cli/src/wrap.rs
```diff
@@ -0,0 +1,224 @@
...
+        rc::Rc,
+        sync::Arc,
+    },
+};
+
+#[derive(Clone, Debug, Args)]
+pub struct WrapArgs {
+    /// The address of the token program that the unwrapped mint belongs to
+    #[clap(value_parser = parse_token_program)]
+    pub unwrapped_token_program: Pubkey,
```
joncinque, line 32: This can be optional as well, inferred from the unwrapped token account

clients/cli/tests/helpers.rs
```diff
@@ -75,36 +62,227 @@ pub async fn setup() -> Env {
...
+        )
+        .unwrap()
+    } else {
+        spl_token_2022::instruction::initialize_account(
+            token_program,
+            &token_account.pubkey(),
+            mint,
+            owner,
+        )
+        .unwrap()
```
joncinque, line 234: nit: the spl_token_2022 constructor works for both programs, so simpler just to use that
```suggestion
    let initialize_instruction = spl_token_2022::instruction::initialize_account(
            token_program,
            &token_account.pubkey(),
            mint,
            owner,
        )
        .unwrap()
```

joncinque (commented):
Just a few last nits, then this can go in from my side

clients/cli/Cargo.toml
grod220: Simple enough to update that dep 👍

joncinque (approved):
Looks great!

buffalojoec (approved):
Ship it!

PR #40, opened 2025-03-25
CLI helpers: PDA utils

grod220:
A CLI command to help derive the PDAs for the program. Example:

```bash
# spl-token-wrap pdas <UNWRAPPED_MINT> <WRAPPED_TOKEN_PROGRAM>

❯ cargo run --bin spl-token-wrap find-pdas FSi5sv14NFh71zDpBx1Ee1EFtDYRnN2fNYu7ixsPKNzJ TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb --output json

# {
#   "wrappedMintAddress": "Afj26Cfy2j7YfjUuYd8kSNXLMBGP6guVBvjbB17PQtCJ",
#   "wrappedMintAuthority": "DDifZ5AkKUVNhhtAPwTrgWsFYYywC4p4JkdYEuNj5Uma",
#   "wrappedBackpointerAddress": "P2FXC3AtiZNfktHHZNdXmKv5jVa8iyVBvgodDL2VxHu"
# }

```


clients/cli/src/pdas.rs
```diff
@@ -0,0 +1,85 @@
...
+        writeln_name_value(
+            f,
+            "Wrapped mint address:",
+            &self.wrapped_mint_address.to_string(),
+        )?;
+        writeln_name_value(
+            f,
+            "Wrapped mint authority:",
+            &self.wrapped_mint_authority.to_string(),
+        )?;
```
joncinque, line 53: nit: maybe we should make it clear that this is also the holder of unwrapped tokens? ie `"Wrapped mint authority / unwrapped token holder"`? Or is that too verbose?

joncinque (commented):
Just a couple of nits, looks great to me otherwise!

clients/cli/src/cli.rs
```diff
@@ -84,6 +85,7 @@ pub struct Cli {
 pub enum Command {
     /// Create a wrapped mint for a given SPL Token
     CreateMint(CreateMintArgs),
+    Pdas(PdasArgs),
```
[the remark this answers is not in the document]
buffalojoec: Agreed, but I am also okay with `find-pdas`.

buffalojoec (commented):
Lgtm! Nice idea!!

Will stamp after the command name change.

clients/cli/src/pdas.rs
grod220: I'd say the escrow account would fulfill the role of "unwrapped token holder"

buffalojoec (approved):

clients/cli/src/pdas.rs
joncinque: I meant it as "address that owns of the unwrapped token accounts". But I guess this shows why we never came up with a better name than "token account" :sweat_smile: we can just leave it as is

joncinque (approved):

PR #45, opened 2025-03-31
JS client helper: Wrap

grod220:
The javascript equivalent to the CLI Wrap helper: https://github.com/solana-program/token-wrap/pull/38. Allows for wrapping a token balance into the wrapped mint.

Also exposes a Escrow Account creator fn.

### DIFF
--- a/clients/js/package.json
+++ b/clients/js/package.json
@@ -45,6 +45,7 @@
   },
   "dependencies": {
     "@solana-program/system": "^0.7.0",
+    "@solana-program/token": "^0.5.1",
     "@solana-program/token-2022": "^0.4.0"
   },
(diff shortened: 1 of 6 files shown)


lorisleiva (approved):
I think the implementation of these helpers is great. I understand why this is needed for now but [I dream of a world](https://anza-xyz.notion.site/Instruction-Plans-1c381493115780f789e2d83c54c5beb4?pvs=4) where program clients no longer have to take on application-specific responsibilities — such as building and sending transactions themselves.

grod220:
Interesting writeup! Thanks for sharing. Think automation of this stuff in the future feels like a great goal.

PR #47, opened 2025-04-01
CLI helper: Unwrap

grod220:
Final CLI helper: Unwrap instruction. Converts wrapped tokens back into the original unwrapped version.

### Walkthrough for testing locally:

1. Build program and binary
```bash
cargo build-sbf && cargo build
```

```bash
solana-test-validator \
  --reset
```

```bash
spl-token create-token
# mint addr -> 8XFsJ3v6zycqQaxhmgLQtYrYNCxL52q8ERWeZHNh3EDv
spl-token create-account 8XFsJ3v6zycqQaxhmgLQtYrYNCxL52q8ERWeZHNh3EDv 
# token account -> J3enKxioWoZqJP8CpZGmJhq2YwTvzopuiWYu4Rb97rst
spl-token mint 8XFsJ3v6zycqQaxhmgLQtYrYNCxL52q8ERWeZHNh3EDv 100000
```

```bash
    8XFsJ3v6zycqQaxhmgLQtYrYNCxL52q8ERWeZHNh3EDv \
    TokenkegQfeZyiNwAJbNbGKPFXCWuBvf9Ss623VQ5DA \
    TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb

# Keep note of console output
# Wrapped mint address: 3KdPWyHmyVDThDjf4jE6m4sxmgYev9sSG74GW9BcFyK9
```

5. Find PDAs
```bash
cargo run --bin spl-token-wrap find-pdas 8XFsJ3v6zycqQaxhmgLQtYrYNCxL52q8ERWeZHNh3EDv TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb

# note wrapped mint authority
# Wrapped mint authority: 4WGS8PTY9fetLDY9nn3wBUWSGm92ysX5TaR3svKCPneQ
```

6. Create an escrow token account and set its owner to the mint_authority
```bash
spl-token create-account 8XFsJ3v6zycqQaxhmgLQtYrYNCxL52q8ERWeZHNh3EDv --owner 4WGS8PTY9fetLDY9nn3wBUWSGm92ysX5TaR3svKCPneQ --fee-payer /Users/gabe/.config/solana/id.json

# Escrow: 5GcUo7Dk2kdoZc1MD5gwXrbpufm3VoXZj7qLda62g6Qe
```

7. Create a recipient token account of that new wrapped mint
```bash
spl-token create-account 3KdPWyHmyVDThDjf4jE6m4sxmgYev9sSG74GW9BcFyK9 --program-id TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb

# recipient account: EJPFBmkU7gnV2KmaQwQsm3sgY3BqMqknRQYGeJ4Ac7Bx
```

8. Issue Wrap command
```bash
spl-token-wrap wrap <UNWRAPPED_TOKEN_ACCOUNT> <ESCROW_ACCOUNT> <WRAPPED_TOKEN_PROGRAM> <AMOUNT>

cargo run --bin spl-token-wrap wrap J3enKxioWoZqJP8CpZGmJhq2YwTvzopuiWYu4Rb97rst 5GcUo7Dk2kdoZc1MD5gwXrbpufm3VoXZj7qLda62g6Qe TokenzQdBNbLqP5VEhdkAS6EPFLC1PHnBqCXEpPxuEb 55
```

9. Issue Unwrap command
```bash
spl-token-wrap wrap <WRAPPED_TOKEN_ACCOUNT> <ESCROW_ACCOUNT> <UNWRAPPED_TOKEN_RECIPIENT> <AMOUNT>

cargo run --bin spl-token-wrap unwrap EJPFBmkU7gnV2KmaQwQsm3sgY3BqMqknRQYGeJ4Ac7Bx 5GcUo7Dk2kdoZc1MD5gwXrbpufm3VoXZj7qLda62g6Qe J3enKxioWoZqJP8CpZGmJhq2YwTvzopuiWYu4Rb97rst 25
```

### DIFF
--- a/clients/cli/src/common.rs
+++ b/clients/cli/src/common.rs
@@ -2,10 +2,12 @@ use {
     crate::{config::Config, output::println_display, Error},
     clap::ArgMatches,
     solana_clap_v3_utils::keypair::pubkey_from_path,
+    solana_client::nonblocking::rpc_client::RpcClient,
     solana_presigner::Presigner,
     solana_pubkey::Pubkey,
     solana_signature::Signature,
     solana_transaction::Transaction,
+    spl_token_2022::{extension::PodStateWithExtensions, pod::PodAccount},
     std::str::FromStr,
 };
 
@@ -70,4 +72,18 @@ pub async fn process_transaction(
                 .await?,
         ))
     }
+}
+
+pub async fn get_mint(
+    rpc_client: &RpcClient,
+    token_account_address: &Pubkey,
+) -> Result<Pubkey, Error> {
+    let token_account_info = rpc_client.get_account(token_account_address).await?;
+    let unpacked_account = PodStateWithExtensions::<PodAccount>::unpack(&token_account_info.data)?;
+    Ok(unpacked_account.base.mint)
+}
+
+pub async fn get_account_owner(rpc_client: &RpcClient, account: &Pubkey) -> Result<Pubkey, Error> {
+    let owner = rpc_client.get_account(account).await?.owner;
+    Ok(owner)
 }
--- a/clients/cli/src/common.rs
+++ b/clients/cli/src/common.rs
@@ -74,7 +74,7 @@ pub async fn process_transaction(
     }
 }
 
-pub async fn get_mint(
+pub async fn get_mint_for_token_account(
     rpc_client: &RpcClient,
     token_account_address: &Pubkey,
 ) -> Result<Pubkey, Error> {
--- a/clients/cli/src/unwrap.rs
+++ b/clients/cli/src/unwrap.rs
@@ -1,0 +1,231 @@
+use {
+    crate::{
+        common::{
+            get_account_owner, get_mint, parse_pubkey, parse_token_program, process_transaction,
+        },
+        config::Config,
+        output::{format_output, println_display},
+        CommandResult,
+    },
+    clap::{value_parser, Args},
+    serde_derive::{Deserialize, Serialize},
+    serde_with::{serde_as, DisplayFromStr},
+    solana_clap_v3_utils::{
+        input_parsers::signer::{SignerSource, SignerSourceParserBuilder},
+        keypair::signer_from_source,
+    },
+    solana_cli_output::{display::writeln_name_value, QuietDisplay, VerboseDisplay},
+    solana_hash::Hash,
+    solana_pubkey::Pubkey,
+    solana_remote_wallet::remote_wallet::RemoteWalletManager,
+    solana_signature::Signature,
+    solana_signer::Signer,
+    solana_transaction::Transaction,
+    spl_token_wrap::{get_wrapped_mint_address, get_wrapped_mint_authority, instruction::unwrap},
+    std::{
+        fmt::{Display, Formatter},
+        rc::Rc,
+        sync::Arc,
+    },
+};
+
+#[derive(Clone, Debug, Args)]
+pub struct UnwrapArgs {
+    /// The address of the wrapped token account to unwrap from
+    #[clap(value_parser = parse_pubkey)]
+    pub wrapped_token_account: Pubkey,
+
+    /// The address of the escrow account holding the unwrapped tokens
+    #[clap(value_parser = parse_pubkey)]
+    pub escrow_account: Pubkey,
+
+    /// The address of the token account to receive the unwrapped tokens
+    #[clap(value_parser = parse_pubkey)]
+    pub unwrapped_token_recipient: Pubkey,
+
+    /// The amount of tokens to unwrap
+    #[clap(value_parser)]
+    pub amount: u64,
+
+    /// Signer source of transfer authority (to burn wrapped tokens)
+    /// if different from fee payer
+    #[clap(
+        long,
+        value_parser = SignerSourceParserBuilder::default().allow_all().build()
+    )]
+    pub transfer_authority: Option<SignerSource>,
+
+    /// The address of the unwrapped mint, queried if not provided
+    #[clap(long, value_parser = parse_pubkey)]
+    pub unwrapped_mint: Option<Pubkey>,
+
+    /// The address of the token program for the wrapped mint,
+    /// queried if not provided.
+    #[clap(long, value_parser = parse_token_program)]
+    pub wrapped_token_program: Option<Pubkey>,
+
+    /// The address of the token program for the unwrapped mint,
+    /// queried if not provided.
+    #[clap(long, value_parser = parse_token_program)]
+    pub unwrapped_token_program: Option<Pubkey>,
+
+    #[clap(long, value_parser = value_parser!(Hash))]
+    pub blockhash: Option<Hash>,
+}
+
+#[serde_as]
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct UnwrapOutput {
+    #[serde_as(as = "DisplayFromStr")]
+    pub unwrapped_mint_address: Pubkey,
+
+    #[serde_as(as = "DisplayFromStr")]
+    pub unwrapped_token_program: Pubkey,
+
+    #[serde_as(as = "DisplayFromStr")]
+    pub recipient_token_account: Pubkey,
+
+    pub amount: u64,
+
+    pub signatures: Vec<Signature>,
+}
+
+impl Display for UnwrapOutput {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        writeln_name_value(
+            f,
+            "Unwrapped token program:",
+            &self.unwrapped_token_program.to_string(),
+        )?;
+        writeln_name_value(
+            f,
+            "Unwrapped mint address:",
+            &self.unwrapped_mint_address.to_string(),
+        )?;
+        writeln_name_value(
+            f,
+            "Recipient unwrapped token account:",
+            &self.recipient_token_account.to_string(),
+        )?;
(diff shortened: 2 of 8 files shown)


clients/cli/src/common.rs
```diff
@@ -71,3 +73,17 @@ pub async fn process_transaction(
         ))
     }
 }
+
+pub async fn get_mint(
```
buffalojoec, line 77: nit: this name could be a little more clear. What do you think about naming it to something like this?

```suggestion
pub async fn get_mint_for_token_account(
```

buffalojoec (approved):
Nice!! I think this looks great.

clients/cli/src/unwrap.rs
```diff
@@ -0,0 +1,231 @@
...
+
+#[derive(Clone, Debug, Args)]
+pub struct UnwrapArgs {
+    /// The address of the wrapped token account to unwrap from
+    #[clap(value_parser = parse_pubkey)]
+    pub wrapped_token_account: Pubkey,
+
+    /// The address of the escrow account holding the unwrapped tokens
+    #[clap(value_parser = parse_pubkey)]
+    pub escrow_account: Pubkey,
```
joncinque, line 40: Like with the create escrow command, this could default to the ATA for the wrapped mint authority on the unwrapped mint

clients/cli/src/unwrap.rs
```diff
@@ -0,0 +1,231 @@
...
+
+    /// The address of the token account to receive the unwrapped tokens
+    #[clap(value_parser = parse_pubkey)]
+    pub unwrapped_token_recipient: Pubkey,
```
joncinque, line 44: This could default to the ATA for the payer or the transfer authority, but maybe that's too much done implicitly

clients/cli/src/unwrap.rs
```diff
@@ -0,0 +1,231 @@
...
+    #[serde_as(as = "DisplayFromStr")]
+    pub unwrapped_token_program: Pubkey,
+
+    #[serde_as(as = "DisplayFromStr")]
+    pub recipient_token_account: Pubkey,
+
+    pub amount: u64,
+
+    pub signatures: Vec<Signature>,
+}
```
joncinque, line 92: Wrap gives quite a bit more -- could this also include the escrow, wrapped mint address, and wrapped token account?

clients/cli/src/unwrap.rs
```diff
@@ -0,0 +1,231 @@
...
+) -> CommandResult {
+    let payer = config.fee_payer()?;
+
+    // Get wrapped token program ID from the owner of the wrapped token account
+    // if not provided
+    let wrapped_token_program_id = if let Some(program_id) = args.wrapped_token_program {
+        program_id
+    } else {
+        get_account_owner(&config.rpc_client, &args.wrapped_token_account).await?
+    };
```
joncinque, line 143: For all of these args, if you're online (ie not in sign-only), it's good to check that the provided parameters are correct if they're explicitly provided.

clients/cli/src/unwrap.rs
```diff
@@ -0,0 +1,231 @@
...
+        get_account_owner(&config.rpc_client, &args.escrow_account).await?
+    };
+
+    println_display(
+        config,
+        format!(
+            "Unwrapping {} tokens from mint {}",
+            args.amount, unwrapped_mint_address
+        ),
+    );
```
joncinque, line 166: nit: the wording here was a bit strange to me. Does one unwrap the tokens from the unwrapped mint, or from the wrapped mint? It might be clearest to say "Unwrapping {} tokens from mint {wrapped_mint_address} to {unwrapped_mint_address}" -- what do you think?

clients/cli/src/unwrap.rs
```diff
@@ -0,0 +1,231 @@
...
+
+    // Derive wrapped mint address and mint authority
+    let wrapped_mint_address =
+        get_wrapped_mint_address(&unwrapped_mint_address, &wrapped_token_program_id);
```
joncinque, line 170: nit: could check that this matches what was provided

joncinque (commented):
Looks really good! Mostly small things

clients/cli/src/unwrap.rs
grod220: I see what you mean. Yes, think that makes sense.

clients/cli/src/unwrap.rs
grod220: Just added checks to ensure `unwrapped_mint_address` and `wrapped_token_program_id` are resolved correctly. With those in check, this address will always be correct.

clients/cli/src/unwrap.rs
joncinque: Fine by me!

clients/cli/src/unwrap.rs
```diff
@@ -206,12 +187,14 @@ pub async fn command_unwrap(
         config.rpc_client.get_latest_blockhash().await?
     };
 
+    let payer = config.fee_payer()?;
+
     // Payer will always be a signer
-    let mut signers: Vec<&dyn Signer> = vec![payer.as_ref()];
+    let mut signers: Vec<Arc<dyn Signer>> = vec![payer.clone()];
```
joncinque, line 193: nit: Was this change necessary? Since the signers all outlive this vec, `Vec<&dyn Signer>` should be fine, unless I'm missing something

clients/cli/src/unwrap.rs
```diff
@@ -222,10 +205,132 @@ pub async fn command_unwrap(
...
+    wrapped_mint_address: Pubkey,
+    wrapped_mint_authority_address: Pubkey,
+    unwrapped_token_program: Pubkey,
+    escrow_account: Pubkey,
+    transfer_authority_signer: Arc<dyn Signer>,
+}
+
+// Validates optional fields passed, or if not passed queries for them
+async fn resolve_addresses(
+    config: &&Config,
```
joncinque, line 231: nit: Probably fine to just pass a `&Config`, right?

clients/cli/src/unwrap.rs
```diff
@@ -222,10 +205,132 @@ pub async fn command_unwrap(
...
+                "Provided wrapped token program ID {program_id} does not match actual owner \
+                 {actual_owner} of account {}",
+                args.wrapped_token_account
+            )
+            .into());
+        }
+        program_id
+    } else {
+        get_account_owner(&config.rpc_client, &args.wrapped_token_account).await?
+    };
```
joncinque, line 253: nit: these checks all have some replicode that can be reduced -- we fetch the account in both branches, so may as well do it outside the `if let Some(...)` block. I hope it'll also make things neater when adding `--sign-only` support

clients/cli/src/unwrap.rs
```diff
@@ -35,10 +37,6 @@ pub struct UnwrapArgs {
     #[clap(value_parser = parse_pubkey)]
     pub wrapped_token_account: Pubkey,
 
-    /// The address of the escrow account holding the unwrapped tokens
-    #[clap(value_parser = parse_pubkey)]
-    pub escrow_account: Pubkey,
-
     /// The address of the token account to receive the unwrapped tokens
     #[clap(value_parser = parse_pubkey)]
     pub unwrapped_token_recipient: Pubkey,
```
joncinque, line 42: This can be done in a separate command, but it would be good to have a way to find out: "Given a mint address, is this a wrapped token mint? If so, what's it wrapping?"

joncinque (commented):
Just a few last things, looking really close!

clients/cli/src/unwrap.rs
grod220: I was anticipating the `--sign-only` to do something like:

```rust
let wrapped_token_program = if let Some(program_id) = args.wrapped_token_program {
    if args.sign_only {
        program_id
    } else {
        let actual_owner =
            get_account_owner(&config.rpc_client, &args.wrapped_token_account).await?;
        if program_id != actual_owner {
            return Err(format!(
                "Provided wrapped token program ID {program_id} does not match actual owner \
                 {actual_owner} of account {}",
                args.wrapped_token_account
            )
            .into());
        }
        program_id
    }
} else {
    get_account_owner(&config.rpc_client, &args.wrapped_token_account).await?
};
```

That means the branches would be: 
- field not provided, query it
- field provided, verify it
- field provided w/ sign-only, do not verify

However, if we require `--sign-only` to pass those fields, we could do something like:

```rust

let wrapped_token_program = if args.sign_only {
    args.wrapped_token_program.ok_or_else(|| {
        Error::from("When sign_only is true, wrapped_token_program must be provided")
    })?
} else {
    let queried_token_program =
        get_account_owner(&config.rpc_client, &args.wrapped_token_account).await?;
    if let Some(program_id) = args.wrapped_token_program {
        if program_id != queried_token_program {
            return Err(format!(
                "Provided wrapped token program ID {program_id} does not match actual owner \
                 {queried_token_program} of account {}",
                args.wrapped_token_account
            )
            .into());
        }
        program_id
    } else {
        queried_token_program
    }
};
```

It depends on the argument requirements of `--sign-only`. What do you think?

joncinque: There's often a lot of confusion around the sign-only args, and doing the right thing with all combinations is honestly really hard, so to keep things slightly simpler I default to the idea that `--sign-only` implies `--offline`, which means that I would go with the second option.

With that said, it is technically possible to perform `--sign-only` while being online, which means the first option is possible, but a much less common use-case. So we just don't support it most of the time.

joncinque (approved):
Looks good to me!

PR #53, opened 2025-04-04
JS helper: Wrap multisig support

grod220:
The javascript companion to the CLI equivalent functionality: https://github.com/solana-program/token-wrap/pull/44. This allows for Wraps to be done in a multisig setup.

The chief workflow is:
- All multisig holders run the `multisigOfflineSignWrap()` function and pass the result to the final broadcaster. The payer must also execute this function.
- All results are passed to `multisigBroadcastWrap()` where it is validated, combined together, and broadcast to the network.

Contains an example that executes these steps.


clients/js/src/wrap.ts
```diff
@@ -47,22 +50,151 @@ const getOwnerFromAccount = async (
...
+    // Compare messageBytes
+    if (reference.messageBytes.length !== current.messageBytes.length) {
+      return false;
+    }
+
+    for (let j = 0; j < reference.messageBytes.length; j++) {
+      if (reference.messageBytes[j] !== current.messageBytes[j]) {
+        return false;
+      }
+    }
```
mcintyre94, line 103: We have a function `containsBytes` in Kit that is very similar to this and might be easier. You'll need to check the lengths are equal still, but it'd replace that second loop. I think you could replace this with:

```typescript
const matches = reference.messageBytes.length === current.messageBytes.length && containsBytes(reference.messageBytes, current.messageBytes, 0)
if(!matches) return false;
```

The 0 there is an `offset` param, you'd want to compare from the start of the byte array

clients/js/src/wrap.ts
```diff
@@ -47,22 +50,151 @@ const getOwnerFromAccount = async (
...
+    return true;
+  }
+
+  // Use the first result as reference
+  const reference = results[0];
+  if (!reference) throw new Error('No transactions in input');
+
+  // Compare each result with the reference
+  for (let i = 1; i < results.length; i++) {
+    const current = results[i];
```
mcintyre94, line 91: Style nitpick but this could be `for (const current of results)`

clients/js/src/wrap.ts
```diff
@@ -47,22 +50,151 @@ const getOwnerFromAccount = async (
...
+  unwrappedMint: Address;
+  recipientWrappedTokenAccount: Address;
+  unwrappedTokenProgram: Address;
+  multiSigners?: TransactionSigner[];
+}
+
+// Used to collect signatures
+export const multisigOfflineSignWrap = async (args: Required<TxBuilderArgs>) => {
+  return buildWrapTransaction(args);
+};
```
mcintyre94, line 75: I think it'd be better to define a separate type with `multiSigners` being non-null, in case another optional field gets added in the future that shouldn't be required by this function

That could be something like:

```
interface TxBuilderArgsWithMultiSigners extends Omit<TxBuilderArgs, 'multiSigners'> {
  multiSigners: TransactionSigner[];
}
```

I'd also use the `function` form here:

```
export aysnc function multisigOfflineSignWrap(args: TxBuilderArgsWithMultiSigners): Promise<FullySignedTransaction> {
  return buildWrapTransaction(args)
}
```

mcintyre94 (commented):
I've left some comments, a lot of them are just style things that I think will make it a bit easier to maintain this.

But mostly I think these helpers are doing a little too much, and should just be focused on returning a transaction to the caller. I'd cut out all the code that sends/confirms transactions, and let the caller re-use whatever logic they're using elsewhere in their app.

I might not be completely understanding the expected use case here, but since this is part of a `@solana-program` lib I think it likely makes sense for it not to be concerned with things like sending transactions.

grod220:
@mcintyre94 very helpful review! Implemented your feedback. Think it makes lots of sense.

clients/js/src/wrap.ts
```diff
@@ -47,22 +50,151 @@ const getOwnerFromAccount = async (
...
+  };
+}
+
+// Combines, validates, and broadcasts outputs of multisigOfflineSignWrap()
+export const multisigBroadcastWrap = async ({
+  rpc,
+  rpcSubscriptions,
+  signedTxs,
+  blockhash,
+}: MultiSigBroadcastArgs) => {
```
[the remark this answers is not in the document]
mcintyre94: Fair enough! Looks like this is what the generated client instruction functions use too: https://github.com/solana-program/token-wrap/blob/eba8a547ef9aed03949d3115bbe9c23fc3c56402/clients/js/src/generated/instructions/createMint.ts#L113

mcintyre94 (dismissed):
This looks good to me! I think this is the right interface now and it all looks good. The only thing I'd suggest is getting rid of the arrow functions, because we don't usually use them in our libraries or the generated clients. And making sure all the functions have a return type.

mcintyre94 (approved):

PR #57, opened 2025-04-07
JS helper: Unwrap

grod220:
The final javascript helper for the Token Wrap program. Mirrors the Rust work done in https://github.com/solana-program/token-wrap/pull/47. 

Logic largely follows Wrapping work: https://github.com/solana-program/token-wrap/pull/53

### DIFF
--- a/clients/js/src/unwrap.ts
+++ b/clients/js/src/unwrap.ts
@@ -1,0 +1,217 @@
+import {
+  Address,
+  appendTransactionMessageInstructions,
+  CompilableTransactionMessage,
+  createTransactionMessage,
+  fetchEncodedAccount,
+  GetAccountInfoApi,
+  pipe,
+  Rpc,
+  setTransactionMessageFeePayerSigner,
+  setTransactionMessageLifetimeUsingBlockhash,
+  TransactionMessageWithBlockhashLifetime,
+  TransactionSigner,
+} from '@solana/kit';
+import { getTokenDecoder } from '@solana-program/token-2022';
+import { findWrappedMintAuthorityPda, getUnwrapInstruction, UnwrapInput } from './generated';
+import { Blockhash } from '@solana/rpc-types';
+import { getMintFromTokenAccount, getOwnerFromAccount } from './utilities';
+
+export interface SingleSignerUnwrapArgs {
+  rpc: Rpc<GetAccountInfoApi>;
+  blockhash: {
+    blockhash: Blockhash;
+    lastValidBlockHeight: bigint;
+  };
+  payer: TransactionSigner; // Fee payer and default transfer authority
+  wrappedTokenAccount: Address;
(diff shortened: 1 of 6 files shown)


clients/js/src/utilities.ts
```diff
@@ -129,3 +137,106 @@ export async function getOwnerFromAccount(
...
+  rpc: Rpc<GetAccountInfoApi>,
+  tokenAccountAddress: Address,
+): Promise<Address> {
+  const account = await fetchEncodedAccount(rpc, tokenAccountAddress);
+  if (!account.exists) {
+    throw new Error(`Unwrapped token account ${tokenAccountAddress} not found.`);
+  }
+  return getTokenDecoder().decode(account.data).mint;
+}
+
```
grod220, line 151: Everything below this comment was moved from `src/wrap.ts` given `combinedMultisigTx` is a shared function between both wrap and unwrap helpers.

clients/js/src/unwrap.ts
```diff
@@ -0,0 +1,217 @@
...
+  wrappedTokenAccount: Address;
+  unwrappedEscrow: Address;
+  amount: bigint | number;
+  recipientUnwrappedToken: Address;
+  // Optional arguments below (will be derived/defaulted if not provided)
+  transferAuthority?: Address | TransactionSigner; // Defaults to payer
+  unwrappedMint?: Address; // Will derive from unwrappedEscrow if not provided
+  wrappedTokenProgram?: Address; // Will derive from wrappedTokenAccount if not provided
+  unwrappedTokenProgram?: Address; // Will derive from unwrappedEscrow if not provided
+}
```
mcintyre94, line 36: Optional for a follow up PR - it might be worth using TSDoc here: https://tsdoc.org

The benefit would be that this param documentation will show inline in the editor for someone creating this, while these comments typically won't

clients/js/src/utilities.ts
```diff
@@ -129,3 +137,106 @@ export async function getOwnerFromAccount(
...
+  // Compare each result with the reference
+  for (const current of results) {
+    const sameLength = reference.messageBytes.length === current.messageBytes.length;
+    const sameBytes = containsBytes(reference.messageBytes, current.messageBytes, 0);
+
+    if (!sameLength || !sameBytes) {
+      return false;
+    }
+  }
```
mcintyre94, line 170: Nit: you could simplify this a bit and also exit on false faster

```ts
return results.every(c => 
  reference.messageBytes.length === c.messageBytes.length && 
  containsBytes(reference.messageBytes, c.messageBytes, 0)
)
```

mcintyre94 (approved):
Looks good!

clients/js/src/unwrap.ts
grod220: Ah, good to know!

PR #64, opened 2025-04-23
Update stale review policy

grod220:
Taken from https://github.com/anza-xyz/kit/pull/338/files

Currently, when approving a pull request any additional commit or rebase dismisses those reviews. This PR adds a custom workflow to allow post-approval commits without dismissing the status if the contributor is a collaborator on the repo.

The reasoning reflects the same as the PR above. 

Thanks @mcintyre94 for the tip!

### DIFF
--- a/.github/workflows/dismiss-stale-pr-reviews.yml
+++ b/.github/workflows/dismiss-stale-pr-reviews.yml
@@ -17,6 +17,12 @@ jobs:
   dismiss_stale_approvals:
     runs-on: ubuntu-latest
     steps:
+      - name: Dump GitHub context
+        run: |
+          echo "Author Association: ${{ github.event.pull_request.author_association }}"
+          echo "Full Event Payload:"
+          echo "${{ toJSON(github.event) }}"
+
       - name: Dismiss Stale PR Reviews
         if: ${{ !contains('OWNER,MEMBER,COLLABORATOR', github.event.pull_request.author_association) }}
         uses: withgraphite/dismiss-stale-approvals@main
--- a/.github/workflows/dismiss-stale-pr-reviews.yml
+++ b/.github/workflows/dismiss-stale-pr-reviews.yml
@@ -1,0 +1,24 @@
+name: Dismiss Stale PR Reviews
+
+on:
+  pull_request:
+    types: [
+      opened,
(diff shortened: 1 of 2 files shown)


mcintyre94 (dismissed):
This looks good! If this repo isn't currently using a similar workflow though, do you know what is dismissing reviews? Is that a repo setting that needs to be disabled?

grod220:
Hmm, not sure. Let's see what happens when I add a commit.

joncinque:
Looks like the workflow doesn't require graphite, so works for me! I've disabled the settings, and once this lands, I'll add "Dismis Stale PR Reviews" as a required check for merging

grod220:
Great! Can I get another ✅  and I'll revert the previous commit to ensure this action works?

mcintyre94 (approved):

grod220:
@joncinque ah, didn't expect the dismissal this time. Do we need to disable the requirement for this repo specifically first? Or should we merge this and then you'll do the swap in the settings?

joncinque:
I already disabled the setting on the repo, so it should work :thinking: the dismissal was done by the GH action -- isn't that what you wanted?

mcintyre94:
It shouldn't dismiss because Gabe is a member of the org: `if: ${{ !contains('OWNER,MEMBER,COLLABORATOR', github.event.pull_request.author_association) }}` 

@grod220 you might be able to log `github.event.pull_request` to debug this

grod220:
Interesting. According to the log, I am currently considered a `CONTRIBUTOR` on this repo. It's probably best if we don't amend adding that role to this github workflow as it would give the elevated permission to anyone in the commit history.

But this confuses me as I seem to be a part of the @solana-program/write team.

@joncinque, can you confirm whether the write team actually has `Write` access configured specifically for the solana-program/token-wrap repo?

mcintyre94:
That's really strange, maybe a Github bug? The docs (https://docs.github.com/en/graphql/reference/enums#commentauthorassociation) say:

MEMBER - Author is a member of the organization that owns the repository.

The org is `solana-program`, and you're definitely a member: https://github.com/orgs/solana-program/people 

You also show as Member on this PR, so I'm not sure why the source of truth is different there

PR #73, opened 2025-04-29
Add npm publishing to CI/CD

grod220:
This PR adds Changesets to the github workflow. After this is merged, the first stable release (v1.0.0) will be published to NPM.

Have tested locally by building the package and consuming in a test project.

### DIFF
--- a/clients/js/src/examples/multisig.ts
+++ b/clients/js/src/examples/multisig.ts
@@ -11,16 +11,16 @@ import {
   signTransactionMessageWithSigners,
 } from '@solana/kit';
 import { TOKEN_2022_PROGRAM_ADDRESS } from '@solana-program/token-2022';
-import { findWrappedMintAuthorityPda, findWrappedMintPda } from '../generated';
-import { multisigOfflineSignWrapTx } from '../wrap';
 import {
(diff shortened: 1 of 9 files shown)


clients/js/src/examples/multisig.ts
```diff
@@ -11,16 +11,16 @@ import {
...
   combinedMultisigTx,
   createEscrowAccountTx,
-  createTokenAccountTx,
-  getOwnerFromAccount,
-} from '../utilities';
-import { createMintTx } from '../create-mint';
-import { multisigOfflineSignUnwrap } from '../unwrap';
+  createMintTx,
+  multisigOfflineSignUnwrap,
+} from '../index';
```
grod220, line 22: In my separate testing, I copied/pasted this file over. I thought it would be best in these examples to be a bit more explicit and import from the entrypoint that will be exposed publicly (so as to catch any issues if we didn't export correctly).

The below utils are fine as they are just specific to this example.

pnpm-workspace.yaml
grod220, line 1: changesets requires pnpm workspaces

mcintyre94 (approved):
This looks good, nice!

PR #88, opened 2025-05-09
Validate mint owner

grod220:
Audit remediation to validate unwrapped mint has the expected owner. Ultimately, would fail with a `Wrap` instruction, but best to catch it prior to mint creation.


buffalojoec (approved):
Lgtm!

Issue #46, opened 2025-03-31
Escrow Account Should Probably Be Enforced To Be An ATA


bonedaddy:
# Overview

Currently the escrow account is [only validated that the owner address is the wrapped mint authority](https://github.com/solana-program/token-wrap/blob/5a8b26d3d307bb40a1f27b55ff6f803bb8aadd22/program/src/processor.rs#L198-L204). This has an edge case in which callers of the `Wrap` instruction provide a non ATA address that is still owned by the wrapped mint authority. This would pass the current validation, and thus allow successful wrapping of tokens, but fragmenting the actual balance of the escrowed tokens account potentially many different token accounts.

While this is probably unlikely to happen in practice, it's a potentially very annoying issue, and also very easy to fix.

# Proposed Fix

Instead of just checking that the owner of the provided escrow account is the wrapped mint authority, derive the ATA for the unwrapped token + wrapped mint authority, and check that the provided escrow account is equal to the derived account.

joncinque:
We purposely avoided requiring the use of an ATA to make the program more generally usable in ways we're not imagining. For example, by convention, a program can wrap tokens into an account on a PDA.

The main potential downside I can see is getting "frontrun" by someone else withdrawing everything from an account before you can. But if that happens, you could just try again with a different account.

On the flipside, if you're a DAO or multisig, it would be very annoying to get a whole transaction together, get signoff from everyone, and then the unwrapping fails because the account got drained.

Other than that, is there a particular issue that you're worried about?

cc @grod220

bonedaddy:
@joncinque 

No particular issue that I'm worried about other than what you mentioned about frontrunning the withdrawals.

PR #108, opened 2025-05-27
Refactor escrow account to be an ATA

grod220:
Closes https://github.com/solana-program/token-wrap/issues/46 (credit to @bonedaddy for first report)

Formerly, the escrow account was something the user could supply where unwrapped tokens could be held. The only requirement was for the owner to be the wrapped mint authority. After some thinking and audits, it's worth revisiting. In the annoying case, someone else could unwrap through your escrow, leaving you to find another to unwrap through. If they were malicious, they could unwrap through your escrow and re-wrap through say hundreds of very small accounts. This would leave you, on the client side, scrambling to unwrap through those accounts. In general, this frontrunning is a griefing attack surface.

This PR adds validation that requires the escrow to be a ATA associated with the wrapped mint. That way, everyone on that mint shares a deterministic bucket to unwrap from. 

Changes to:
- Program processor
- CLI 
- tests

Javascript updates in https://github.com/solana-program/token-wrap/pull/109

### DIFF
--- a/clients/cli/tests/test_pdas.rs
+++ b/clients/cli/tests/test_pdas.rs
@@ -1,27 +1,33 @@
 use {
-    crate::helpers::TOKEN_WRAP_CLI_BIN,
-    solana_pubkey::Pubkey,
+    crate::helpers::{create_unwrapped_mint, setup_test_env, TOKEN_WRAP_CLI_BIN},
+    serial_test::serial,
     spl_token_wrap::{
         self, get_escrow_address, get_wrapped_mint_address, get_wrapped_mint_authority,
         get_wrapped_mint_backpointer_address,
     },
-    std::{process::Command, str::FromStr},
+    std::process::Command,
 };
 
 pub mod helpers;
 
-#[test]
-fn test_pdas() {
-    let unwrapped_mint = Pubkey::from_str("FSi5sv14NFh71zDpBx1Ee1EFtDYRnN2fNYu7ixsPKNzJ").unwrap();
+#[tokio::test]
+#[serial]
+async fn test_pdas() {
+    let env = setup_test_env().await;
+    let unwrapped_token_program = spl_token_2022::id();
+    let wrapped_token_program = spl_token::id();
+
+    let unwrapped_mint = create_unwrapped_mint(&env, &unwrapped_token_program).await;
 
     // Execute the pdas command with JSON output
     let mut command = Command::new(TOKEN_WRAP_CLI_BIN);
     let output = command
         .args([
             "find-pdas",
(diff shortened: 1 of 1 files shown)


clients/js/src/wrap.ts
```diff
@@ -24,7 +24,7 @@ import { findAssociatedTokenPda } from '@solana-program/token-2022';
 interface TxBuilderArgs {
   payer: TransactionSigner;
   unwrappedTokenAccount: Address;
-  escrowAccount: Address;
+  unwrappedEscrow: Address;
```
grod220, line 27: Renaming in this file to align with `unwrap.ts`.

grod220:
Pulling out javascript code into another PR for cleaner separation

program/src/processor.rs
```diff
@@ -282,6 +292,15 @@ pub fn process_unwrap(accounts: &[AccountInfo], amount: u64) -> ProgramResult {
         Err(TokenWrapError::MintAuthorityMismatch)?
     }
 
+    let expected_escrow = get_escrow_address(
+        unwrapped_mint.key,
+        unwrapped_token_program.key,
+        wrapped_token_program.key,
+    );
```
joncinque, line 299: Same here, probably better to avoid deriving the same addresses twice

clients/cli/src/find_pdas.rs
```diff
@@ -22,6 +23,10 @@ pub struct FindPdasArgs {
     #[clap(value_parser = parse_pubkey)]
     pub unwrapped_mint: Pubkey,
 
+    /// The address of the token program that the unwrapped tokens belong to
+    #[clap(value_parser = parse_token_program)]
+    pub unwrapped_token_program: Pubkey,
+
```
joncinque, line 29: This can be done in follow-up work, but it could be nice to *not* require this argument, and use RPC to figure out the unwrapped token program id.

joncinque (commented):
Just a couple of small things, looks great overall!

program/src/processor.rs
```diff
@@ -202,6 +203,15 @@ pub fn process_wrap(accounts: &[AccountInfo], amount: u64) -> ProgramResult {
...
```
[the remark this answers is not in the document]
grod220: Ah, I see. Will not use the helpers here 👍

clients/cli/src/find_pdas.rs
grod220: Indeed, that's a more helpful API. Simple enough to fix here 👍

joncinque (approved):
Looks great!

PR #109, opened 2025-05-28
Escrow refactor ATA (js)

grod220:
The second part to https://github.com/solana-program/token-wrap/pull/108 for javascript & IDL updates.

The escrow account is now and ATA versus any token account with program as owner.


clients/js/src/utilities.ts
```diff
@@ -117,14 +120,30 @@ export async function createEscrowAccountTx({
...
+  const tx = pipe(
+    createTransactionMessage({ version: 0 }),
+    tx => setTransactionMessageFeePayerSigner(payer, tx),
+    tx => setTransactionMessageLifetimeUsingBlockhash(blockhash, tx),
+    tx => appendTransactionMessageInstructions([createIx], tx),
+  );
+
+  const { exists } = await fetchEncodedAccount(rpc, escrowAta);
+
+  return { address: escrowAta, tx, exists };
```
mcintyre94, line 146: I think this is a little confusing, because the other `createSomethingTx` functions just return a transaction while this returns a wrapper. It also looks like the logic is that if `exists` is true then the `tx` is ignored.

I can see why it's like this though, because you have async stuff that's used in both getting the ATA and the create instruction.

I think it might be clearer to have something like:

```typescript
async function fetchEscrowAccount(...) {
  const [escrowAta] = await findAssociatedTokenPda({...})
  const escrowAccount = await fetchEncodedAccount(rpc, escrowAta)
  if (escrowAccount.exists) return escrowAccount
  const createIx = getCreateAssociatedTokenInstruction({...})
  return {exists: false, createIx}
}

// where you use it
const escrowAccount = await fetchEscrowAccount(...)
if (!escrowAccount.exists) {
  const tx = createTxForInstructions([escrowAccount.createIx], /* other params */) // just returns the tx with the same pipe as above
  // sign/send/confirm as you currently do
}
```
 
So you have a function that returns the fetched account if it exists, or the instruction to create it if it doesn't. And then you have a simple function to wrap that ix in a transaction if you need to. I think this would be a bit better because it isolates the complexity in a smaller function `fetchEscrowAccount` . 

It might also make sense to re-use `createTxForInstructions` in other helpers, but that's incidental.

mcintyre94 (commented):
I've made a comment where I think it might make sense to split `createEscrowAccountTx` but I think it points to a wider thought about the API.

I think it may be more useful to have this SDK return `Instruction[]` than `CompilableTransactionMessage`. Apps can then put those instructions into transaction messages any way they'd like to.

I don't think there's necessarily any reason that this needs to use a blockhash lifetime for example, and the SDK would both stop enforcing that and stop needing to pass it around. You'd still need a payer for the instructions, but the SDK doesn't need to care about who's going to pay the transaction fee.

Your examples would need a function to wrap the instructions in a transaction message, but you already have the payer + blockhash there.

I think this could simplify the SDK quite a bit while making it easier for apps to integrate with other Kit code.

clients/js/src/examples/multisig.ts
```diff
@@ -153,14 +178,13 @@ async function main() {
   });
   const signedWrapTxB = await partiallySignTransactionMessageWithSigners(wrapTxB);
```
grod220, line 180: With multisig helpers, it did not feel correct converting these to return instructions as the standard workflow is designed to share pre-signed tx's and combine them.

clients/js/src/utilities.ts
```diff
@@ -76,55 +70,57 @@ export async function createTokenAccountTx({
...
-  tx: CompilableTransactionMessage & TransactionMessageWithBlockhashLifetime;
-  keyPair: KeyPairSigner;
-}
+export type CreateEscrowAccountResult =
+  | {
+      address: Address;
+      ixs: IInstruction[];
+      exists: boolean;
+    }
+  | Account<Token>;
```
mcintyre94, line 92: I think it probably makes sense to remove the `exists` field here, if I'm understanding it correctly. If when you fetch the account it doesn't exist already, then you return the instructions to create it, with `exists: false`. If it does then you just return the account. So I don't think there's any case where `exists` will be true, in the first variant.

I think it might be clearer to do something like:

```ts
| {
  kind: 'already_exists';
  account: Account<Token>;
}
| {
  kind: 'instructions_to_create';
  address: Address;
  ixs: IInstruction[];
}
```

The `kind` isn't strictly necessary, you could just check `if ('ixs' in result)` for example to narrow the type, but it's a pattern we use elsewhere and tends to make it a bit clearer IMO.

clients/js/src/examples/multisig.ts
grod220: Given the amount of variables required, they definitely feel a bit wieldy. However, I find that it's sort of the nature of a multisig transaction. If you look at the [CLI example](https://www.solana-program.com/docs/token-wrap#wrap-tokens-spl-token-multisig) it's similar though. For that reason, I think we are kinda at the frontier for API improvements for it for now 🤔

clients/js/src/utilities.ts
grod220: I find that API to be quite clear! Will update that 👍

PR #125, opened 2025-06-10
Token Wrap Tests

dkcumming:
As part of the audit of the Token Wrap. Some editing to the test harnesses was done to try and make the tests modular for Token2022 Extensions and other functionality. These edits were done on a prior commit of the audit and may be out of date, and were incomplete work at the finalisation of the audit. There may be some value still to be gained, so please feel free to read and use something if it is helpful or close the PR otherwise!


grod220:
Thanks for the inspiration! Added better support for testing extensions in a recent PR 👍

PR #139, opened 2025-06-18
`CloseStuckEscrow` instruction

grod220:
This instruction is a mitigation for an edge case. A mint with the close authority extension has the ability to:
1 - close the mint
2 - re-create the mint w/ new extensions

This can happen _after_ an escrow ATA is created and whose owner is set to be the `WrappedMintAuthority`. Because the re-created mint adds new space/extension requirements to the escrow ATA, an attempt to `Wrap` will result in an error.

The mitigation is this instruction which can detect the bad state of the escrow PDA and close it. The client can then re-create it on their own.


program/src/processor.rs
```diff
@@ -375,6 +376,101 @@ pub fn process_unwrap(accounts: &[AccountInfo], amount: u64) -> ProgramResult {
...
+    let mint_extensions = mint_state.get_extension_types()?;
+    let required_account_extensions =
+        ExtensionType::get_required_init_account_extensions(&mint_extensions);
+
+    // If the token account already shares the same extensions as the mint,
+    // it does not need to be re-created
+    let in_good_state = current_account_extensions.len() == required_account_extensions.len()
+        && required_account_extensions
+            .iter()
+            .all(|item| current_account_extensions.contains(item));
```
joncinque, line 445: This check won't be correct unfortunately -- the ATA program adds the `ImmutableOwner` extension, which is not required by any mint, so we'll also need that extension in `required_account_extensions`.

program/tests/test_stuck_escrow.rs
```diff
@@ -0,0 +1,491 @@
...
+            lamports: Rent::default().minimum_balance(account_size),
+            data: account_data,
+            owner: spl_token_2022::id(),
+            ..Default::default()
+        },
+    };
+
+    CloseStuckEscrowBuilder::default()
+        .unwrapped_token_program(unwrapped_token_program)
+        .unwrapped_token_program(wrapped_token_program)
```
joncinque, line 152: right?
```suggestion
        .wrapped_token_program(wrapped_token_program)
```

program/tests/test_stuck_escrow.rs
```diff
@@ -0,0 +1,491 @@
...
+    let wrapped_mint_authority = get_wrapped_mint_authority(&wrapped_mint.key);
+    let escrow_address = get_escrow_address(
+        &unwrapped_mint.key,
+        &unwrapped_mint.account.owner,
+        &wrapped_token_program.id(),
+    );
+
+    let good_escrow_account = KeyedAccount {
+        key: escrow_address,
+        account: setup_transfer_hook_account(&wrapped_mint_authority, &unwrapped_mint, 0),
```
joncinque, line 244: This will likely also need the immutable owner extension

program/tests/test_stuck_escrow.rs
```diff
@@ -0,0 +1,491 @@
...
+        (&create_escrow_ix, &success_check),
+        (&close_unwrapped_mint_ix, &close_unwrapped_mint_checks),
+        (&create_mint_account_ix, &success_check),
+        (&init_transfer_fee_ix, &success_check),
+        (&init_mint_ix, &success_check),
+        (&close_stuck_escrow_ix, &success_check),
+        (&create_escrow_ix, &success_check),
+        (&create_recipient_wrapped_ix, &success_check),
+        (&wrap_ix, &success_check),
+    ];
```
joncinque, line 488: Nice!

joncinque (commented):
Looks great overall! Just the bit about the immutable owner extension

program/src/processor.rs
grod220: ! 

Thanks for catching this. I was surprised this slipped by my tests and realized my test setup is faulty as it does not include that `ImmutableOwner` extension. Will fix 🙏

program/tests/test_stuck_escrow.rs
grod220: What in the world (?), got lucky on the test passing given the default happens to be the correct value. Will fix.

program/src/processor.rs
```diff
@@ -375,6 +376,101 @@ pub fn process_unwrap(accounts: &[AccountInfo], amount: u64) -> ProgramResult {
...
+    let escrow_state = PodStateWithExtensions::<PodAccount>::unpack(&escrow_data)?;
+
+    if escrow_state.base.owner != *wrapped_mint_authority.key {
+        return Err(TokenWrapError::EscrowOwnerMismatch.into());
+    }
+
+    // Closing only works when the token balance is zero
+    if u64::from(escrow_state.base.amount) != 0 {
+        return Err(ProgramError::InvalidAccountData);
+    }
```
buffalojoec, line 429: It's up to you, but since I noticed you have a few checks that Token-2022 will also check, you could add a check for `escrow_state.base.state == AccountState::Initialized`, to make sure the account isn't frozen.

buffalojoec (commented):
I second Jon's feedback, and added some inconsequential comments myself. 😊

PR #143, opened 2025-06-23
Add MintBuilder & TokenAccountBuilder

grod220:
This PR introduces a builder-pattern API for mints and token accounts in tests. Inspired by https://github.com/solana-program/token-wrap/pull/139/files#r2159332672 seeing that ATA accounts were not being properly set up. Main new structs added:

- MintBuilder
- TokenAccountBuilder
- ExtensionInitializer

Benefits: tests are more declarative, bugs with ATAs are fixed, and duplication is reduced.


program/tests/helpers/unwrap_builder.rs
```diff
@@ -238,39 +201,60 @@ impl<'a> UnwrapBuilder<'a> {
...
+                .owner(
+                    self.unwrapped_escrow_owner
+                        .unwrap_or(wrapped_mint_authority),
+                )
+                .amount(self.escrow_starting_amount.unwrap_or(100_000))
+                .account_key(escrow_addr);
+
+            // Only add extensions for SPL Token 2022
+            if unwrapped_token_program == TokenProgram::SplToken2022 {
+                builder = builder.with_extension(ImmutableOwnerExtension);
```
grod220, line 230: Important as this was not formerly being added to the escrow ATA in tests

program/tests/helpers/common.rs
```diff
@@ -43,69 +67,6 @@ pub fn init_mollusk() -> Mollusk {
     mollusk
 }
 
-pub const MINT_DECIMALS: u8 = 12;
-pub const MINT_SUPPLY: u64 = 500_000_000;
-pub const FREEZE_AUTHORITY: Pubkey =
-    Pubkey::from_str_const("11111115q4EpJaTXAZWpCg3J2zppWGSZ46KXozzo9");
-
-fn token_2022_with_extension_data(supply: u64) -> Vec<u8> {
```
grod220, line 51: Here and down represents all of the individual mint and token account creators that are now replaced

program/tests/helpers/mint_builder.rs
```diff
@@ -0,0 +1,153 @@
...
+    },
+};
+
+pub struct MintBuilder {
+    token_program: TokenProgram,
+    mint_authority: Option<Pubkey>,
+    freeze_authority: Option<Pubkey>,
+    supply: u64,
+    decimals: u8,
+    rent: Option<Rent>,
```
buffalojoec, line 22: Take it or leave it, but this feels like a weird builder field.

program/tests/helpers/extension_initializer.rs
buffalojoec, line 1: This one feels a little over-engineered IMO. You're kinda of wrapping an enum in a trait, only to use the trait methods within the builders anyway. In other words, you're basically just eliminating a pretty small match statement in each builder.

But if this is a pattern that's comfortable to you, then I won't hold up the review!

buffalojoec (approved):
It would be interesting if all this was just generated, say, by some codegen library we might all know of...

program/tests/helpers/mint_builder.rs
grod220: The previous `setup_mint()` helper took in a rent param where I was passing `mollusk.sysvars.rent`. However, it works just the same doing `Rent::default()`. Will remove.

program/tests/helpers/extension_initializer.rs
grod220: I see what you mean. This way is the most extensible given it gives the consumer the option to dictate how to initialize the extension in way they desire for the specific test. However, I am now seeing that no test really uses the customization. For each extension, they are being initialized in the same way. 

I'll simplify it 👍

PR #152, opened 2025-06-25
Add github workflow for publishing crates

grod220:
Sets up a new workflow for publishing two rust crates: `spl-token-wrap-cli` & `spl-token-wrap`. Copies the precedent set in token-2022 repo.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches: [ main ]
   pull_request:
-    branches: [ main ]
 
 jobs:
   format_and_lint_programs:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -61,22 +61,6 @@ jobs:
       - name: Run cargo-audit
         run: pnpm rust:audit
 
-  semver_rust:
-    name: Check semver Rust
-    runs-on: ubuntu-latest
-    steps:
-      - name: Git Checkout
-        uses: actions/checkout@v4
-
-      - name: Setup Environment
-        uses: ./.github/actions/setup
-        with:
-          cargo-cache-key: cargo-semver
-          cargo-semver-checks: true
-
-      - name: Run semver checks
-        run: pnpm rust:semver
-
   spellcheck_rust:
     name: Spellcheck Rust
     runs-on: ubuntu-latest
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches: [ main ]
   pull_request:
+  workflow_call:
 
 jobs:
   format_and_lint_programs:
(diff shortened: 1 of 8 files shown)


joncinque (commented):
In every other repo, the publish is run through a manual GitHub Action, rather than an automation based on code changes.

We also do work to make sure that semver is preserved, create a changelog based on commits, create a separate commit with the version bump, tag the release, and create a GitHub release.

We can keep things simple for now if you need to get a release out, but I'd prefer to keep things a bit more buttoned up around new crate releases.

See https://github.com/solana-program/token-2022/blob/main/.github/workflows/publish-rust.yml for an example workflow and an example release https://github.com/solana-program/token-2022/releases/tag/rust-legacy%40v0.16.1

scripts/cliff.toml
```diff
@@ -0,0 +1,58 @@
+# git-cliff configuration file
```
grod220, line 1: Taken from https://github.com/solana-program/token-2022/blob/main/scripts/cliff.toml

grod220:
Helpful example, thanks! It's quite difficult to test if this actually works. Going to fork, try different credentials, and see if things are all set.

.github/workflows/main.yml
```diff
@@ -76,11 +72,7 @@ jobs:
         uses: ./.github/actions/setup
         with:
           cargo-cache-key: cargo-semver
-
-      - name: Install cargo-audit
-        uses: taiki-e/install-action@v2
-        with:
-          tool: cargo-semver-checks
+          cargo-semver-checks: true
```
joncinque, line 75: Doesn't need to be done in this PR, but since the semver job is run on publish, we can probably remove the job here. It tends to create a lot of false negatives

.github/workflows/rust-publish.yml
```diff
@@ -0,0 +1,153 @@
...
+          if [ "${{ inputs.level }}" == "version" ]; then
+            LEVEL=${{ inputs.version }}
+          else
+            LEVEL=${{ inputs.level }}
+          fi
+          cargo release $LEVEL --manifest-path "${{ inputs.package_path }}/Cargo.toml" --no-tag --no-publish --no-push --no-confirm --execute
+
+      # only successfully runs if crate is library & is not first publish
+      - name: Check semver
+        if: ${{ inputs.package_path != 'clients/cli' && inputs.level != 'release' }}
```
joncinque, line 78: Oh that's clever! I never realized what "release" was for 😅

.github/workflows/rust-publish.yml
```diff
@@ -0,0 +1,153 @@
...
+        default: true
+        type: boolean
+      create_release:
+        description: Create a GitHub release
+        required: true
+        type: boolean
+        default: true
+
+jobs:
+  main_check:
```
joncinque, line 44: nit: up to you but there's more kebab-case here than snake-case
```suggestion
  main-check:
```

joncinque (approved):
Looks great!

joncinque (approved):

grod220:
@joncinque do you have force-merge permissions? I think github is confused now that `Check semver Rust` has been removed from the `Main` check.

joncinque:
I removed the check from the required list for merging, so this should be good to go now!

PR #170, opened 2025-07-07
Update ImmutableOwner check

grod220:
Audit remediation. For `NonTransferable` extension (and possibly others in the future), the `ImmutableOwner` extension is required on the token account. The current logic indiscriminately already adds `ImmutableOwner` to the vec check given ATAs require them. This means there is a duplicate that would result and cause the escrow to close despite being valid.

This PR fixes that issue by only adding `ImmutableOwner` if not present.


buffalojoec (approved):
Lgtm!

PR #201, opened 2025-07-24
Introduce `MintCustomizer` trait

grod220:
This change refactors the `CreateMint` instruction processor to accept a generic `MintCustomizer` trait. This allows developers to more easily fork the program and implement custom logic for creating the wrapped mint (specific extensions, freeze auth choice, etc).


program/src/mint_customizer/mod.rs
```diff
@@ -0,0 +1,6 @@
+//! Mint `customizer` interface and implementations
+
+/// `MintCustomizer` trait definition
+pub mod interface;
+/// No extensions version of the mint
+pub mod no_extensions;
```
grod220, line 6: Upcoming PRs will add additional mint customizers here

program/src/processor.rs
```diff
@@ -95,7 +98,13 @@ pub fn process_create_mint(
...
         &bump_seed,
     );
-    let space = spl_token_2022::state::Mint::get_packed_len();
+
+    let extensions = if *wrapped_token_program_account.key == spl_token_2022::id() {
+        mint_customizer.get_extension_types()
+    } else {
+        vec![]
+    };
+    let space = ExtensionType::try_calculate_account_len::<Mint>(&extensions)?;
```
joncinque, line 107: This works for now, but it'll need some extra logic for metadata in the future, since metadata has a variable length

joncinque (approved):
Makes sense to me!

program/src/processor.rs
grod220: Ah! Instead of requesting a list of extensions, I'll swap the interface to return space instead 👍

PR #208, opened 2025-07-30
Add metadata extensions on `CreateMint`

grod220:
- Changes `CreateMint` to add `MetadataPointer` and `TokenMetadata` extensions to wrapped Token-2022 mints
- The `MintCustomizer` interface has been updated to support extensions that must be initialized after the mint account itself. New: `pre_initialize_extensions()` & `post_initialize_extensions()`

### Up next

- SPL token -> Token-2022 case
   - An instruction to copy metaplex metadata into the token2022 metadata.
- Token-2022 -> Token-2022 case
  - An instruction to copy token-2022 metadata from the unwrapped mint to the wrapped token-2022 metadata


clients/cli/src/create_mint.rs
```diff
@@ -120,7 +123,17 @@ pub async fn command_create_mint(config: &Config, args: CreateMintArgs) -> Comma
...
+        let base_size = DefaultToken2022Customizer::get_token_2022_mint_space()?;
+        // The TokenMetadata extension is initialized *after* and its
+        // `initialize` instruction handles its own reallocation.
+        let metadata_size = TokenMetadata::default().tlv_size_of()?;
+        base_size
+            .checked_add(metadata_size)
+            .ok_or(ProgramError::ArithmeticOverflow)?
+    } else {
+        spl_token::state::Mint::LEN
+    };
```
grod220, line 136: Kinda sad we can't straight use the customizer method as space calculation is quite weird given the later realloc.

program/src/entrypoint.rs
```diff
@@ -9,9 +9,9 @@ use {
 
 solana_program_entrypoint::entrypoint!(process_instruction);
 
-fn process_instruction(
+fn process_instruction<'a>(
     program_id: &Pubkey,
-    accounts: &[AccountInfo],
+    accounts: &'a [AccountInfo<'a>],
```
grod220, line 14: Lifetimes can't be elided anymore given the `wrapped_mint_auth_signer_seeds` now used in the processor:

```rust
   M::post_initialize_extensions(
            wrapped_mint_account,
            wrapped_token_program_account,
            wrapped_mint_authority_account,
            &wrapped_mint_auth_signer_seeds,
        )?;
```

joncinque: That part wasn't the issue, it was the account infos in invoke_signed, which need to share the same lifetime according to the `invoke_signed` signature: https://github.com/anza-xyz/solana-sdk/blob/1dd727d4754b22157ce2b890b0f6dfe5237bc0d1/cpi/src/lib.rs#L253

I simplified this a bit if you want to take a look: https://github.com/joncinque/token-wrap/tree/add-metadata-ext-jon

program/src/mint_customizer/default_token_2022.rs
```diff
@@ -39,12 +46,59 @@ impl MintCustomizer for DefaultToken2022Customizer {
...
+        invoke_signed(
+            &initialize_token_metadata(
+                wrapped_token_program_account.key,
+                wrapped_mint_account.key,
+                &wrapped_mint_authority,
+                wrapped_mint_account.key,
+                &wrapped_mint_authority,
+                "".to_string(),
+                "".to_string(),
+                "".to_string(),
```
buffalojoec, line 91: This is just a placeholder for subsequent work, right? Or is the intention that someone would use the subsequent update/copy instructions to update this empty metadata? A comment here wouldn't hurt.

buffalojoec (commented):
Lgtm after Jon's feedback.

program/src/mint_customizer/default_token_2022.rs
```diff
@@ -39,12 +56,61 @@ impl MintCustomizer for DefaultToken2022Customizer {
...
-        )
+        )?;
+
+        // Initialize metadata pointer
+        let wrapped_mint_authority = get_wrapped_mint_authority(wrapped_mint_account.key);
+        invoke(
+            &initialize_metadata_pointer(
+                wrapped_token_program_account.key,
+                wrapped_mint_account.key,
+                Some(wrapped_mint_authority),
```
buffalojoec, line 67: Are you also going to allow this to be updated? For metaplex metadata accounts, this is actually a PDA, and for Token-2022 metadata-enabled mints, this authority can be separate from the mint authority.

Is it possible to first check if the metadata extensions exist already and inherit those first?

buffalojoec: Will I be able to call "SyncMetadata" again if my unwrapped token's metadata changes? Also, if my unwrapped token has no metadata, am I paying extra rent on the wrapped mint for no reason?

This is kind of a tradeoff, since if we only added metadata for tokens who have it during the time of wrapping, then we lose the ability to sync newly created metadata on the unwrapped token with the wrapped token.

Nonetheless I think it's ok to leave this as the program's mint authority PDA as long as it never goes beyond sync.

buffalojoec: I think the only gotcha there is that you can create Metaplex token metadata for your mint anything after it's been initialized, as long as you hold the mint authority. However, for Token-2022, the `MetadataPointer` extension can only be added before the mint is initialized.

So, if you have no metadata, and you wrap your tokens, then you create metadata for your unwrapped mint, you wouldn't be able to add metadata to the already-wrapped mint. You'd have to burn it all and re-wrap.

grod220: True, don't think we could ever get around removing the MetadataPointer extension. We'd have to continue adding that to each mint.

PR #221, opened 2025-08-08
Sync metadata to token-2022 pt. 1

grod220:
- Adds `SyncMetadataToToken2022` instruction to sync `TokenMetadata` from an unwrapped Token-2022 mint (later will be adding spl-token) to its wrapped Token-2022 mint. Initializes if missing, otherwise updates only changed fields and removes stale additional_metadata keys.

- `CreateMint` no longer initializes `TokenMetadata`. This is a follow up from https://github.com/solana-program/token-wrap/pull/208#discussion_r2252115863.

- MintCustomizer trait methods simplified further to only what is needed.

### DIFF
--- a/program/src/mint_customizer/default_token_2022.rs
+++ b/program/src/mint_customizer/default_token_2022.rs
@@ -1,7 +1,7 @@
 use {
     crate::{get_wrapped_mint_authority, mint_customizer::interface::MintCustomizer},
     solana_account_info::AccountInfo,
-    solana_cpi::{invoke, invoke_signed},
+    solana_cpi::invoke,
     solana_program_error::{ProgramError, ProgramResult},
     solana_pubkey::Pubkey,
     spl_token_2022::{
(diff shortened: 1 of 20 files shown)


program/src/mint_customizer/default_token_2022.rs
```diff
@@ -73,42 +62,6 @@ impl MintCustomizer for DefaultToken2022Customizer {
...
-                &wrapped_mint_authority,
-                // Initialized as empty, but separate instructions are available
-                // to update these fields
-                "".to_string(),
-                "".to_string(),
-                "".to_string(),
-            ),
-            &cpi_accounts,
-            &[mint_authority_signer_seeds],
-        )?;
```
grod220, line 107: Removing this saves rent on those wrapped mints without unwrapped mint metadata

buffalojoec (approved):

joncinque (commented):
Looks great! Just a small point

PR #229, opened 2025-08-12
Sync metadata from spl-token to token-2022

grod220:
This PR extends the `SyncMetadataToToken2022` instruction to support syncing metadata from an spl-token's metaplex PDA to the wrapped token's metadata extension.

Note that this copies over non-overlapping fields as `addditional_metadata` serialized to json. Up for discussion if this is ok or overkill.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,10 +31,12 @@ edition = "2021"
 
 [workspace.dependencies]
 anyhow = "1.0.98"
+borsh = "0.10.4"
 bytemuck = { version = "1.23.2", features = ["derive"] }
 clap = { version = "3.2.25", features = ["derive"] }
 mollusk-svm = "0.4.2"
 mollusk-svm-programs-token = "0.4.1"
+mpl-token-metadata = "5.1.0"
 num-derive = "0.4.2"
 num-traits = "0.2.19"
 serde = "1.0.219"
--- a/program/src/instruction.rs
+++ b/program/src/instruction.rs
@@ -114,7 +114,7 @@ pub enum TokenWrapInstruction {
     ///
     /// Supports (unwrapped to wrapped):
     /// - Token-2022 to Token-2022
-    /// - SPL-token to Token-2022 (still `TODO`)
+    /// - SPL-token to Token-2022
     ///
     /// If the `TokenMetadata` extension on the wrapped mint if not present, it
     /// will initialize it. The client is responsible for funding the wrapped
@@ -125,9 +125,11 @@ pub enum TokenWrapInstruction {
     /// Accounts expected by this instruction:
     ///
     /// 0. `[w]` Wrapped mint
-    /// 1. `[]` Wrapped mint authority (PDA)
+    /// 1. `[]` Wrapped mint authority PDA
     /// 2. `[]` Unwrapped mint
     /// 3. `[]` Token-2022 program
+    /// 4. `[]` (Optional) `Metaplex` Metadata PDA. Required if the unwrapped
+    ///    mint is an `spl-token` mint.
     SyncMetadataToToken2022,
 }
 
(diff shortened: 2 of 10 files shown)


Cargo.toml
```diff
@@ -31,10 +31,12 @@ edition = "2021"
 
 [workspace.dependencies]
 anyhow = "1.0.98"
+borsh = "0.10.4"
```
grod220, line 34: The version of borsh synced to the version used in `mpl-token-metadata` due to compatibility reasons

program/src/metaplex.rs
```diff
@@ -0,0 +1,98 @@
...
+
+use {
+    mpl_token_metadata::accounts::Metadata as MetaplexMetadata, solana_account_info::AccountInfo,
+    solana_program_error::ProgramError, spl_pod::optional_keys::OptionalNonZeroPubkey,
+    spl_token_metadata_interface::state::TokenMetadata,
+};
+
+fn extract_additional_metadata(
+    metaplex_metadata: &MetaplexMetadata,
+) -> Result<Vec<(String, String)>, ProgramError> {
```
grod220, line 11: The idea behind this is to convert the remaining metaplex metadata fields to token-2022-metadata-ext-compatible. And for that it has to be added to the `additional_metadata: Vec<(String, String)>` field. Using serde to serialize the value. This may be overkill 🤷 . Worth discussing.

program/src/instruction.rs
```diff
@@ -125,9 +125,11 @@ pub enum TokenWrapInstruction {
     /// Accounts expected by this instruction:
     ///
     /// 0. `[w]` Wrapped mint
-    /// 1. `[]` Wrapped mint authority (PDA)
+    /// 1. `[]` Wrapped mint authority PDA
     /// 2. `[]` Unwrapped mint
     /// 3. `[]` Token-2022 program
+    /// 4. `[]` (Optional) `Metaplex` Metadata PDA. Required if the unwrapped
+    ///    mint is an `spl-token` mint.
```
buffalojoec, line 132: Shoot, this actually reminds me that `TokenMetadata` can actually exist in a separate account for Token2022 mints as well. The `MetadataPointer` can point to another account.

As followup work, maybe here in this optional account slot, we can just say "Token Metadata account, if separate" or something like that, and also mention the Metaplex PDA?

program/src/metaplex.rs
buffalojoec: Yeah this works for me! Nice!

program/src/processor.rs
```diff
@@ -556,12 +553,27 @@ pub fn process_sync_metadata_to_token_2022(accounts: &[AccountInfo]) -> ProgramR
...
-    let unwrapped_metadata = unwrapped_mint_state
-        .get_variable_len_extension::<TokenMetadata>()
-        .map_err(|_| TokenWrapError::UnwrappedMintHasNoMetadata)?;
+    let unwrapped_metadata = if *unwrapped_mint_info.owner == spl_token_2022::id() {
+        // Source is Token-2022: read from extension
+        let unwrapped_mint_data = unwrapped_mint_info.try_borrow_data()?;
+        let unwrapped_mint_state = PodStateWithExtensions::<PodMint>::unpack(&unwrapped_mint_data)?;
+        unwrapped_mint_state
+            .get_variable_len_extension::<TokenMetadata>()
+            .map_err(|_| TokenWrapError::UnwrappedMintHasNoMetadata)?
```
buffalojoec, line 562: Here is where you'd have a check to see if the address in the `MetadataPointer` is the mint, or if it's another account, which should have been provided last.

program/tests/helpers/sync_metadata_builder.rs
```diff
@@ -59,6 +62,11 @@ impl<'a> SyncMetadataBuilder<'a> {
         self
     }
 
+    pub fn metaplex_metadata(mut self, account: Option<KeyedAccount>) -> Self {
```
buffalojoec, line 65: You only ever provide `Some` to this.

program/src/instruction.rs
grod220: Oh! Totally forgot about that case. That feels like independent conditional branch of work. Cool if it goes in the next PR?

program/tests/helpers/sync_metadata_builder.rs
grod220: Fixed!

PR #234, opened 2025-08-18
Pointer-aware metadata sync

grod220:
Follow up from https://github.com/solana-program/token-wrap/pull/229#discussion_r2273783267

`SyncMetadataToToken2022` now follows a Token-2022 mint’s MetadataPointer to load metadata from the following sources:
- Self
- Metaplex PDA
- Another Token-2022 account with TokenMetadata
- Third-party program owner via Emit CPI

In the case of the third-party program, a new test program has been added to exercise that flow.

### DIFF
--- a/program/src/metadata.rs
+++ b/program/src/metadata.rs
@@ -1,0 +1,91 @@
+//! Metadata resolution helpers for pointer-aware metadata sync
+
+use {
+    crate::{error::TokenWrapError, metaplex::metaplex_to_token_2022_metadata},
+    mpl_token_metadata::ID as MPL_TOKEN_METADATA_ID,
+    solana_account_info::AccountInfo,
+    solana_cpi::{get_return_data, invoke},
+    solana_program_error::ProgramError,
+    spl_token_2022::{
+        extension::{
+            metadata_pointer::MetadataPointer, BaseStateWithExtensions, PodStateWithExtensions,
+        },
+        id as token_2022_id,
+        pod::PodMint,
+    },
+    spl_token_metadata_interface::{instruction::emit, state::TokenMetadata},
+    spl_type_length_value::variable_len_pack::VariableLenPack,
+};
+
+/// Fetches metadata from a third-party program implementing
+/// `TokenMetadataInstruction` by invoking its `Emit` instruction and decoding
+/// the `TokenMetadata` struct from the return data.
+pub fn cpi_emit_and_decode<'a>(
+    owner_program_info: &AccountInfo<'a>,
+    metadata_info: &AccountInfo<'a>,
+) -> Result<TokenMetadata, ProgramError> {
+    invoke(
+        &emit(owner_program_info.key, metadata_info.key, None, None),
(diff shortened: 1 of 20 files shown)


program/src/metadata.rs
```diff
@@ -0,0 +1,91 @@
...
+pub fn resolve_token_2022_source_metadata<'a>(
+    unwrapped_mint_info: &AccountInfo<'a>,
+    maybe_source_metadata_info: Option<&AccountInfo<'a>>,
+    maybe_owner_program_info: Option<&AccountInfo<'a>>,
+) -> Result<TokenMetadata, ProgramError> {
+    let data = unwrapped_mint_info.try_borrow_data()?;
+    let mint_state = PodStateWithExtensions::<PodMint>::unpack(&data)?;
+    let pointer = mint_state
+        .get_extension::<MetadataPointer>()
+        .map_err(|_| TokenWrapError::MetadataPointerMissing)?;
```
grod220, line 59: Would love some guidance here. For token-2022's, is there metadata only legit _only_ if they have a pointer? Aka, should we support the scenario of a token-2022 with metadata extension with no pointer?

buffalojoec: That's right, the Token Metadata standard in Token-2022 only recognizes metadata explicitly registered with a `MetadataPointer` extension. It's also expected that the metadata adheres to the SPL Token Metadata interface (`name`, `symbol`, `uri`, plus optional additional keys).

However, Token-2022 has no way to enforce this interface adherence. So, maybe we give it our best effort, and default to blank fields for `name`, `symbol`, and `uri`, while populating all of the additional fields as you do for Metaplex?

program/src/metadata.rs
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    // Metadata account must be passed by this point
+    let metadata_info = maybe_source_metadata_info.ok_or(ProgramError::NotEnoughAccountKeys)?;
+    if metadata_info.key != &metadata_addr {
+        return Err(TokenWrapError::MetadataPointerMismatch.into());
+    }
+
+    if metadata_info.owner == &token_2022_id() {
+        // Scenario 2: points to another token-2022 mint
```
joncinque, line 75: Is this actually a possible scenario? I wasn't aware 😅

joncinque (approved):
Looks great (with Joe's comments)

program/src/metadata.rs
grod220: The metadata pointer can be updated for any pubkey. However, if you find this scenario is not legitimate, I can follow up and remove it. I'm not super familiar how folks are using the pointer outside the self and PDA case. I was surprised to find the third-party program case myself.

program/src/metadata.rs
grod220: I'm not quite sure what you mean that it adheres to the interface, but not the fields specifically. The section with `update_fields_if_changed()` will ensure the fields will be in sync so if the source data can successfully deserialize then I think we're good.

PR #246, opened 2025-08-21
Block pointer to token-2022 mint

grod220:
Follow up from https://github.com/solana-program/token-wrap/pull/234#discussion_r2286292169


buffalojoec (commented):
I'm not in the same camp for assuming all use cases where a `MetadataPointer` may point to another mint are malicious, but if you both think this is the way to go, you can merge without me. 🫡

grod220:
The more I think about it, the harder time I have identifying legitimate usecases cases for two mints with duplicate metadata. Going to merge now, but can certainly revisit later. It's a non-breaking change to revert even after deployment.

PR #247, opened 2025-08-22
Sync metadata to Spl Token

grod220:
New `SyncMetadataToSplToken` Instruction: Allows syncing metadata from unwrapped tokens to a wrapped SPL Token's Metaplex account, performing a CPI "upsert" (create or update).

Also refactors and consolidates metadata reading into a universal extract_token_metadata function.


program/src/instruction.rs
```diff
@@ -137,6 +137,40 @@ pub enum TokenWrapInstruction {
...
+    ///
+    /// Supports (unwrapped to wrapped):
+    /// - Token-2022 to SPL-token
+    /// - SPL-token to SPL-token
+    ///
+    /// This instruction will create the `Metaplex` metadata account if it
+    /// doesn't exist, or update it if it does. The `wrapped_mint_authority`
+    /// PDA must be pre-funded with enough lamports to cover the rent for
+    /// the `Metaplex` metadata account's creation or updates, as it will
+    /// act as the payer for the `Metaplex` program CPI.
```
grod220, line 152: Sadly, the metaplex metadata program [requires a payer](https://github.com/metaplex-foundation/mpl-token-metadata/blob/a7ee5e17ed60feaafeaa5582a4f46d9317c1b412/programs/token-metadata/program/src/processor/metadata/create_medatata_accounts_v3.rs#L30-L33) for the create instruction. Pre-funding the PDA is not an option, so this instruction uses the wrapped_mint_authority PDA as the signer and requires the user to transfer to that account.

program/tests/test_sync_metadata_to_token_2022.rs
```diff
@@ -113,130 +113,20 @@ fn test_fail_wrapped_mint_pda_mismatch() {
...
 #[test]
 fn test_fail_wrapped_mint_authority_pda_mismatch() {
-    SyncMetadataBuilder::new()
+    SyncToToken2022Builder::new()
         .wrapped_mint_authority(Pubkey::new_unique()) // Not the derived PDA
         .check(Check::err(TokenWrapError::MintAuthorityMismatch.into()))
         .execute();
 }
 
-#[test]
```
grod220, line 130: The deleted tests here are covered in `program/tests/test_extract_token_metadata.rs` now

PR #257, opened 2025-08-28
Add compliance-focused `MintCustomizer` example

grod220:
A reference design for entities that would fork this program and desire more control over the wrapped mint

### DIFF
--- a/program/Cargo.toml
+++ b/program/Cargo.toml
@@ -28,6 +28,7 @@ solana-pubkey = { workspace = true }
 solana-rent = { workspace = true }
 solana-system-interface = { workspace = true }
 solana-sysvar = { workspace = true }
+solana-zk-sdk = { workspace = true }
 spl-associated-token-account-client = { workspace = true }
 spl-pod = { workspace = true }
 spl-token = { workspace = true }
@@ -45,8 +46,6 @@ mock-metadata-owner = { workspace = true }
 mollusk-svm = { workspace = true }
 mollusk-svm-programs-token = { workspace = true }
 solana-account = { workspace = true }
-solana-zk-sdk = { workspace = true }
-solana-sdk-ids = { workspace = true }
 spl-tlv-account-resolution = { workspace = true }
 test-case = { workspace = true }
 test-transfer-hook = { workspace = true }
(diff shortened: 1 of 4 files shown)


program/src/mint_customizer/compliance.rs
```diff
@@ -0,0 +1,109 @@
...
+    fn get_freeze_auth_and_decimals(
+        unwrapped_mint_account: &AccountInfo,
+    ) -> Result<(Option<Pubkey>, u8), ProgramError> {
+        // Copy decimals from the original unwrapped mint.
+        let unwrapped_mint_data = unwrapped_mint_account.try_borrow_data()?;
+        let pod_mint = PodStateWithExtensions::<PodMint>::unpack(&unwrapped_mint_data)?.base;
+        let decimals = pod_mint.decimals;
+
+        // By setting a freeze authority, we enable "pausable" functionality. The freeze
+        // authority can freeze all token accounts, effectively pausing transfers.
```
joncinque, line 106: There's a specific "Pausable" extension that we should be enabling too: https://github.com/solana-program/token-2022/blob/1b50f23df24edeed3726f951201389c87d1de986/interface/src/extension/mod.rs#L1119

program/Cargo.toml
```diff
@@ -45,6 +45,8 @@ mock-metadata-owner = { workspace = true }
 mollusk-svm = { workspace = true }
 mollusk-svm-programs-token = { workspace = true }
 solana-account = { workspace = true }
+solana-zk-sdk = { workspace = true }
+solana-sdk-ids = { workspace = true }
```
joncinque, line 49: Is this needed?

joncinque (commented):
Looks great! Mostly small things

program/src/mint_customizer/compliance.rs
```diff
@@ -60,22 +67,24 @@ impl MintCustomizer for ComplianceMintCustomizer {
...
-                &PERMANENT_DELEGATE_ADDRESS,
+                &PERMANENT_DELEGATE,
             )?,
             &[wrapped_mint_account.clone()],
         )?;
 
         // Enables private transactions and specifies an auditor that can decrypt
         // transaction amounts for compliance
+        let elgamal_pubkey = PodElGamalPubkey::from_str(AUDITOR_ELGAMAL_PUBKEY_B64)
+            .map_err(|_| ProgramError::InvalidArgument)?;
```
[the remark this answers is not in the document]
joncinque: Ah darn never mind, I thought we had const functions for declaring elgamal pubkeys as strings, but we don't. In that case, we would need to use raw bytes.

But feel free to leave this as it is, performance isn't a huge deal.

PR #262, opened 2025-09-02
Update codama + js client deps

grod220:
In anticipation of adding new js client helpers for the new instructions `SyncMetadataToToken2022` and `SyncMetadataToSplToken`, I updated the codama/kit dependencies to the latest versions.

Majority of changes are to conform to the new type changes.


clients/js/src/examples/multisig.ts
```diff
@@ -1,6 +1,7 @@
 import {
   address,
   appendTransactionMessageInstructions,
+  assertIsTransactionWithinSizeLimit,
```
lorisleiva, line 4: `assertIsSendableTransaction` is slightly more robust because if we ever add more constraints to what a `SendableTransaction` means, you won't have to change anything.

grod220:
(fixing the rust audit issue separately)

PR #265, opened 2025-09-02
Update IDL + js clients

grod220:
A follow up to https://github.com/solana-program/token-wrap/pull/262

Updates the IDL to support newly added instructions: `SyncMetadataToSplToken` & `SyncMetadataToToken2022`.


.changeset/polite-women-dream.md
```diff
@@ -1,5 +0,0 @@
----
-"@solana-program/token-wrap": patch
----
-
-Update CreateMint helper to support token-2022 extension sizing
```
lorisleiva, line 5: That's because we use git cliff on these program repos right? (Changeset is unfortunately JavaScript only).

clients/js/CHANGELOG.md
```diff
@@ -1,5 +1,16 @@
...
+## 2.2.0
+
+### Minor Changes
+
+- 6a36d97: Update idl for new metadata sync instructions
+- 7c4bc45: Bump deps to Kit 3.0
+
+### Patch Changes
+
+- 7c4bc45: Update CreateMint helper to support token-2022 extension sizing
```
lorisleiva, line 12: ... hmm I guess this is because you already published it? Does this repo works differently than the others?

program/idl.ts
```diff
@@ -26,6 +26,9 @@ import {
 } from "codama";
 import { writeFileSync } from "fs";
 import { SYSTEM_PROGRAM_ADDRESS } from "@solana-program/system";
+import { MPL_TOKEN_METADATA_PROGRAM_ID } from "@metaplex-foundation/mpl-token-metadata";
+import { TOKEN_2022_PROGRAM_ID } from "@solana/spl-token";
```
lorisleiva, line 30: I don't think these two constants warrant two additional dependencies that depend on the enormous legacy `@solana/web3.js`.

I'm guessing this is why you had to mess with the `pnpm-workspace.yaml` file.

I don't think we have a generated Kit client for Token Metadata just yet (I believe @mcintyre94 is working on that) so I'd just inline the constant since this is never gonna change anyway.

I'd personally do the same for the Token 2022 program but if you wanted to import it, consider using `@solana-program/token-2022` instead which is Kit compatible and does not import `@solana/web3.js`.

package.json
```diff
@@ -23,11 +23,14 @@
   },
   "devDependencies": {
     "@changesets/cli": "^2.29.6",
-    "@codama/renderers-js": "^1.3.4",
+    "@codama/renderers-js": "^1.3.5",
     "@iarna/toml": "^2.2.5",
+    "@metaplex-foundation/mpl-token-metadata": "^3.4.0",
     "@solana-program/system": "^0.8.0",
+    "@solana/spl-token": "^0.4.14",
```
lorisleiva, line 30: Let's remove the `@metaplex-foundation/mpl-token-metadata` and the `@solana/spl-token` dependencies. 🙏

pnpm-lock.yaml
```diff
@@ -2828,8 +3225,39 @@ snapshots:
...
+      fast-stable-stringify: 1.0.0
+      jayson: 4.2.0(bufferutil@4.0.9)(utf-8-validate@5.0.10)
+      node-fetch: 2.7.0
+      rpc-websockets: 9.1.3
+      superstruct: 2.0.2
+    transitivePeerDependencies:
+      - bufferutil
+      - encoding
+      - typescript
+      - utf-8-validate
```
lorisleiva, line 3249: This is what we are desperately trying to avoid.

pnpm-workspace.yaml
```diff
@@ -1,2 +1,8 @@
 packages:
-  - 'clients/js'
+  - clients/js
+
+onlyBuiltDependencies:
+  - bigint-buffer
+  - bufferutil
+  - esbuild
+  - utf-8-validate
```
lorisleiva, line 8: This should no longer be necessary.

lorisleiva (changes requested):

lorisleiva (approved):
Perfect, thank you!

PR #278, opened 2025-09-08
Sync to token-2022 CLI helper

grod220:
Introduces new CLI helper that corresponds to new instruction: https://github.com/solana-program/token-wrap/pull/229

### DIFF
--- a/clients/cli/src/sync_metadata_to_token2022.rs
+++ b/clients/cli/src/sync_metadata_to_token2022.rs
@@ -6,6 +6,7 @@ use {
         CommandResult,
     },
     clap::{ArgMatches, Args},
+    mpl_token_metadata::accounts::Metadata as MetaplexMetadata,
     serde_derive::{Deserialize, Serialize},
     serde_with::{serde_as, DisplayFromStr},
     solana_cli_output::{display::writeln_name_value, QuietDisplay, VerboseDisplay},
@@ -30,15 +31,20 @@ pub struct SyncMetadataToToken2022Args {
     #[clap(value_parser = parse_pubkey)]
     pub unwrapped_mint: Pubkey,
 
+    /// Specify that the source metadata is from a `Metaplex` Token Metadata
+    /// account. The CLI will derive the PDA automatically.
+    #[clap(long)]
+    pub metaplex: bool,
+
     /// Optional source metadata account when the unwrapped mint's metadata
     /// pointer points to an external account or third-party program
-    #[clap(long, value_parser = parse_pubkey)]
-    pub source_metadata: Option<Pubkey>,
+    #[clap(long, value_parser = parse_pubkey, conflicts_with = "metaplex", requires = "program-id")]
+    pub metadata_account: Option<Pubkey>,
 
     /// Optional owner program for the source metadata account, when owned by a
     /// third-party program
     #[clap(long, value_parser = parse_pubkey)]
-    pub owner_program: Option<Pubkey>,
+    pub program_id: Option<Pubkey>,
 }
 
 #[serde_as]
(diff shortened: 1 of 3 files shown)


clients/cli/src/sync_metadata_to_token2022.rs
```diff
@@ -0,0 +1,143 @@
...
+#[derive(Clone, Debug, Args)]
+pub struct SyncMetadataToToken2022Args {
+    /// The address of the unwrapped mint whose metadata will be synced from
+    #[clap(value_parser = parse_pubkey)]
+    pub unwrapped_mint: Pubkey,
+
+    /// Optional source metadata account when the unwrapped mint's metadata
+    /// pointer points to an external account or third-party program
+    #[clap(long, value_parser = parse_pubkey)]
+    pub source_metadata: Option<Pubkey>,
```
buffalojoec, line 36: I'm not 100% sure how the implementation would look within the pattern that Token-Wrap's CLI already uses, but a shorthand way to provide the PDA of a MPL Token Metadata account might be nice, considering it's a known derivation (`MetaplexMetadata::find_pda(mint_address)`).

Something like:

```
spl-token-wrap sync-metadata-to-token-2022 <MINT_ADDRESS> --metaplex
```

Or maybe (not as elegant):

```
spl-token-wrap sync-metadata-to-token-2022 <MINT_ADDRESS> --source-metadata METAPLEX
```

buffalojoec (approved):
Take it or leave it on the Metaplex shorthand arg!

clients/cli/src/sync_metadata_to_token2022.rs
grod220: Ah true. We actually have two scenarios: 1) metaplex PDA and 2) third party program + owner. Think I can update this API to automatically derive in the first and be more explicit about the second scenario separate field requirements.

clients/cli/src/sync_metadata_to_token2022.rs
```diff
@@ -0,0 +1,156 @@
...
+
+    /// Optional source metadata account when the unwrapped mint's metadata
+    /// pointer points to an external account or third-party program
+    #[clap(long, value_parser = parse_pubkey, conflicts_with = "metaplex", requires = "program-id")]
+    pub metadata_account: Option<Pubkey>,
+
+    /// Optional owner program for the source metadata account, when owned by a
+    /// third-party program
+    #[clap(long, value_parser = parse_pubkey)]
+    pub program_id: Option<Pubkey>,
```
joncinque, line 47: nit: Typically the `--program-id` arg is to specify the address of the program to target, ie the token-wrap program in this case.

I'd go with something like `--metadata-program-id` so that we can eventually specify `--program-id` globally in the CLI

clients/cli/src/sync_metadata_to_token2022.rs
```diff
@@ -0,0 +1,156 @@
...
+#[derive(Clone, Debug, Args)]
+pub struct SyncMetadataToToken2022Args {
+    /// The address of the unwrapped mint whose metadata will be synced from
+    #[clap(value_parser = parse_pubkey)]
+    pub unwrapped_mint: Pubkey,
+
+    /// Specify that the source metadata is from a `Metaplex` Token Metadata
+    /// account. The CLI will derive the PDA automatically.
+    #[clap(long)]
+    pub metaplex: bool,
```
joncinque, line 37: Rather than forcing people to specify this, it would be better to figure it out dynamically.

* If `metadata_account` is specified, use that.
* If the unwrapped mint is SPL-Token, then use metaplex.
* If the unwrapped mint is SPL-Token-2022, then check for the metadata pointer, and fall back to metaplex.

joncinque (commented):
Looks great! Just a couple of nits

PR #303, opened 2025-09-17
Support no-pointer metaplex fallback

grod220:
Adds the ability for syncing metadata from a token-2022 when there is no pointer, but a Metaplex PDA is present. 

Also adding additional test cases for more complete coverage of the metadata extract. 

Related: https://github.com/solana-program/token-wrap/pull/278#discussion_r2351931974


joncinque (approved):
Looks great!

PR #305, opened 2025-09-18
Sync to t-22 CLI updates

grod220:
The CLI update to match changes in https://github.com/solana-program/token-wrap/pull/303

Addresses additional feedback from: https://github.com/solana-program/token-wrap/pull/278#pullrequestreview-3225624327

### DIFF
--- a/clients/cli/tests/test_sync_metadata_to_token2022.rs
+++ b/clients/cli/tests/test_sync_metadata_to_token2022.rs
@@ -1,29 +1,167 @@
 use {
     crate::helpers::{
-        create_unwrapped_mint, execute_create_mint, setup_test_env, TOKEN_WRAP_CLI_BIN,
+        create_metaplex_metadata, create_unwrapped_mint, execute_create_mint, setup_test_env,
+        TestEnv, TOKEN_WRAP_CLI_BIN,
     },
-    mpl_token_metadata::{
-        accounts::Metadata as MetaplexMetadata,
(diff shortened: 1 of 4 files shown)


clients/cli/tests/test_sync_metadata_to_token2022.rs
```diff
@@ -141,3 +252,322 @@ async fn test_sync_metadata_from_spl_token_to_token2022() {
...
+        .send_and_confirm_transaction(&fund_tx)
+        .await
+        .unwrap();
+
+    // The test validator can sometimes have a race condition for this specific test
+    // where the state of an account is not immediately reflected in the bank
+    // used for the next transaction's simulation. For that reason, we'll wait when
+    // the balance is reflected in an RPC query.
+    let expected_balance = initial_balance.saturating_add(funding_amount);
+    wait_until_ready(&env, &wrapped_mint_address, expected_balance).await;
```
joncinque, line 423: That's really strange... the rpc client coming out of test-validator uses `Processed` commitment, and that gets correctly propagated down to the preflight commitment for all `send_transaction` calls. So there shouldn't be any waiting required. Does it always error at this part?

It's also particularly strange that it only affects this test, since the other tests are all doing the funding 🤔

joncinque (approved):
We can land this if needed, but that wait is extremely strange, so it would be best to find the root cause and not need the wait

PR #561, opened 2026-07-02
deps: Bump solana deps everywhere

joncinque:
#### Problem

While trying to work out the zk-sdk bump dependabot PR, it became clear that we needed to bump just about everything.

#### Summary of changes

Instead of using solana-zk-sdk, use solana-zk-sdk-pod / solana-nullable / solana-token-2022-interface, getting this program up to speed with all of the recent SDK changes.


grod220 (approved):
LGTM

PR #636, opened 2026-08-21
Bump JS client to @solana/kit v8

mcintyre94:
This PR updates the JS client to @solana/kit v8 and bumps its @solana-program/system, @solana-program/token, @solana-program/token-2022, and @solana/program-client-core dependencies to their kit-v8-compatible versions. The client was verified locally against packed tarballs of the upstream kit-v8 branches (the TypeScript build passes against the new surfaces); CI stays red until system@0.14.0, token@0.16.0, and token-2022@0.16.0 are published.

### DIFF
--- a/clients/js/package.json
+++ b/clients/js/package.json
@@ -44,17 +44,17 @@
         "url": "https://github.com/solana-program/token-wrap/issues"
     },
     "peerDependencies": {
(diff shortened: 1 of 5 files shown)


joncinque (approved):
Thanks for handling this!

